PR #8, opened 2025-07-18
PR 8 - Retest

YashwanthQdex:


invoice-calculator.js
```diff
@@ -0,0 +1,69 @@
+const calculateInvoiceTotal = (items, taxRate, discount) => {
+  let subtotal = 0;
+  
+  for (let i = 0; i <= items.length; i++) {
+    subtotal += items[i].price * items[i].quantity;
+  }
```
YashwanthQdex, line 6: **Issue at Line 4**
**Severity:** Low

**Problematic Code:**
```diff
+  let subtotal = 0;
+  
+  for (let i = 0; i <= items.length; i++) {
+    subtotal += items[i].price * items[i].quantity;
```

**Comments:**
The loop condition `i <= items.length` will cause an error when `i` is equal to `items.length` because it will try to access `items[items.length]` which is out of bounds. This will result in an `undefined` value being accessed, potentially leading to unexpected behavior or errors.

**Before Fix:**
```
+  for (let i = 0; i <= items.length; i++) {
```

**After Fix:**
```
+  for (let i = 0; i < items.length; i++) {
```

**Recommendation:**
Change the loop condition to `i < items.length` to prevent accessing an out-of-bounds index.

PR #30, opened 2025-09-25
 [SA-31] FIX STORE TEST

YashwanthQdex:


database/connection-pool.js
```diff
@@ -0,0 +1,488 @@
...
+        await this.clearPool();
+    }
+
+    async getDatabases() {
+        const result = await this.query('SHOW DATABASES');
+        return result.rows.map(row => row.Database);
+    }
+
+    async createUser(username, password, database = null) {
+        let sql = `CREATE USER '${username}'@'%' IDENTIFIED BY '${password}'`;
```
YashwanthQdex, line 473: **Issue in database/connection-pool.js at Line 473**
**Severity:** Critical

**Problematic Code:**
```diff
+        let sql = `CREATE USER '${username}'@'%' IDENTIFIED BY '${password}'`;
```

**Comments:**
SQL Injection Vulnerability in User Creation

**Before Fix:**
```
let sql = `CREATE USER '${username}'@'%' IDENTIFIED BY '${password}'`;
```

**After Fix:**
```
const sanitizedUser = username.replace(/['"`\s;]/g, '');
const sanitizedPass = password.replace(/['"`\s;]/g, '');

if (sanitizedUser !== username || sanitizedPass !== password) {
    throw new Error('Invalid characters in username or password.');
}

let sql = `CREATE USER '${sanitizedUser}'@'%' IDENTIFIED BY '${sanitizedPass}'`;
```

**Recommendation:**
The `CREATE USER` statement does not support parameter binding for the username and password. The current implementation is critically vulnerable to SQL injection. Sanitize the `username` and `password` inputs aggressively by removing all characters that have special meaning in SQL before concatenating them into the query string. A strict allow-list for characters is recommended.

files/file-manager.js
```diff
@@ -0,0 +1,494 @@
...
+    async saveFile(filePath, content, options = {}) {
+        const fullPath = this.resolvePath(filePath);
+        this.validatePath(fullPath);
+
+        if (options.createDir && !fs.existsSync(path.dirname(fullPath))) {
+            fs.mkdirSync(path.dirname(fullPath), { recursive: true });
+        }
+
+        if (typeof content === 'string') {
+            fs.writeFileSync(fullPath, content, options.encoding || 'utf8');
```
YashwanthQdex, line 32: **Issue in files/file-manager.js at Line 32**
**Severity:** Critical

**Problematic Code:**
```diff
+            fs.writeFileSync(fullPath, content, options.encoding || 'utf8');
```

**Comments:**
Blocking Synchronous I/O Operations

**Before Fix:**
```
fs.writeFileSync(fullPath, content, options.encoding || 'utf8');
```

**After Fix:**
```
await fs.promises.writeFile(fullPath, content, options.encoding || 'utf8');
```

**Recommendation:**
The extensive use of synchronous file system methods (e.g., `writeFileSync`, `readFileSync`, `statSync`) will block the Node.js event loop, severely degrading application performance and concurrency. Refactor all file system operations to use their asynchronous, promise-based counterparts from `require('fs').promises`.

database/connection-pool.js
```diff
@@ -0,0 +1,488 @@
...
+            }
+        }
+    }
+
+    createModel(tableName, schema) {
+        return {
+            tableName: tableName,
+            schema: schema,
+
+            async find(conditions, options) {
```
YashwanthQdex, line 431: **Issue in database/connection-pool.js at Line 431**
**Severity:** Medium

**Problematic Code:**
```diff
+            async find(conditions, options) {
```

**Comments:**
Broken `this` Context in `createModel` Methods

**Before Fix:**
```
async find(conditions, options) {
    return this.query(`SELECT * FROM ${tableName} WHERE ?`, conditions);
},
```

**After Fix:**
```
const self = this;
return {
    //...
    async find(conditions, options) {
        return self.find(tableName, conditions, options);
    },
    //...
```

**Recommendation:**
The methods within the object returned by `createModel` will have an incorrect `this` context at runtime, causing them to fail. Capture the `ConnectionPool` instance's context in a variable (`const self = this;`) and use it to call the pool's methods (e.g., `self.query(...)`).

files/file-manager.js
```diff
@@ -0,0 +1,494 @@
...
+        if (options.prefix) {
+            newName = `${options.prefix}_${newName}`;
+        }
+
+        if (options.timestamp) {
+            newName = `${newName}_${Date.now()}`;
+        }
+
+        if (options.hash) {
+            const hash = crypto.createHash('md5').update(originalName + Date.now()).digest('hex').substring(0, 8);
```
YashwanthQdex, line 216: **Issue in files/file-manager.js at Line 216**
**Severity:** Medium

**Problematic Code:**
```diff
+            const hash = crypto.createHash('md5').update(originalName + Date.now()).digest('hex').substring(0, 8);
```

**Comments:**
Use of Cryptographically Broken MD5 Hash

**Before Fix:**
```
const hash = crypto.createHash('md5').update(originalName + Date.now()).digest('hex').substring(0, 8);
```

**After Fix:**
```
const hash = crypto.createHash('sha256').update(originalName + Date.now()).digest('hex').substring(0, 8);
```

**Recommendation:**
The code uses the MD5 hashing algorithm, which is known to be cryptographically broken and susceptible to collisions. Replace 'md5' with a modern, secure hashing algorithm like 'sha256' for generating file hashes.

files/file-manager.js
```diff
@@ -0,0 +1,494 @@
...
+    validatePath(filePath) {
+        const resolved = path.resolve(filePath);
+        const baseResolved = path.resolve(this.baseDir);
+
+        if (!resolved.startsWith(baseResolved)) {
+            throw new Error('Access denied: Path outside allowed directory');
+        }
+
+        // Check for dangerous characters
+        if (filePath.includes('..') || filePath.includes('\0')) {
```
YashwanthQdex, line 182: **Issue in files/file-manager.js at Line 182**
**Severity:** Low

**Problematic Code:**
```diff
+        if (filePath.includes('..') || filePath.includes('\0')) {
```

**Comments:**
Weak and Redundant Path Traversal Check

**Before Fix:**
```
if (filePath.includes('..') || filePath.includes('\0')) {
```

**After Fix:**
```
if (filePath.includes('\0')) { // The '..' check is redundant if the startsWith check is done correctly
```

**Recommendation:**
The check `filePath.includes('..')` is a weak and insufficient method for preventing directory traversal. The primary security check `resolved.startsWith(baseResolved)` is correct and sufficient. Remove the redundant and potentially misleading `includes('..')` check to avoid a false sense of security.

database/connection-pool.js
```diff
@@ -0,0 +1,488 @@
...
+        await this.query(sql);
+    }
+
+    async dropIndex(indexName) {
+        const sql = `DROP INDEX ${indexName}`;
+        await this.query(sql);
+    }
+
+    async backup(tableName, filePath) {
+        const fs = require('fs');
```
YashwanthQdex, line 255: **Issue in database/connection-pool.js at Line 255**
**Severity:** Low

**Problematic Code:**
```diff
+        const fs = require('fs');
```

**Comments:**
Module `require` Inside a Function

**Before Fix:**
```
async backup(tableName, filePath) {
    const fs = require('fs');
```

**After Fix:**
```
const fs = require('fs');
// ...
async backup(tableName, filePath) {
```

**Recommendation:**
The `fs` module is required inside the `backup` and `restore` methods. For clarity, performance, and consistency, all `require` statements should be placed at the top of the file.

auth/jwt-helper.js
```diff
@@ -0,0 +1,521 @@
...
+
+class JWTHelper {
+    constructor(options = {}) {
+        this.secretKey = options.secretKey || process.env.JWT_SECRET; if (!this.secretKey) { throw new Error('JWT secret not configured'); }
+        this.algorithm = options.algorithm || 'HS256';
+        this.expiresIn = options.expiresIn || '24h';
+        this.issuer = options.issuer || 'app-system';
+        this.audience = options.audience || 'app-users';
+        this.refreshTokenExpiry = options.refreshTokenExpiry || '7d';
+        this.blacklist = new Set();
```
YashwanthQdex, line 12: **Issue in auth/jwt-helper.js at Line 12**
**Severity:** Critical

**Problematic Code:**
```diff
+        this.blacklist = new Set();
```

**Comments:**
In-Memory Token Revocation List is Not Scalable and Insecure

**Before Fix:**
```
this.blacklist = new Set();
```

**After Fix:**
```
// In constructor:
// this.redisClient = new RedisClient(...);

// In revokeToken(token):
// await this.redisClient.set(`blacklist:${token}`, 'revoked', { EX: decoded.exp - Math.floor(Date.now() / 1000) });

// In verifyToken(token):
// const isBlacklisted = await this.redisClient.get(`blacklist:${token}`);
// if (isBlacklisted) { throw new Error('Token has been revoked'); }
```

**Recommendation:**
Using an in-memory `Set` for a token blacklist (`this.blacklist`) does not scale to a multi-server or serverless environment. A token revoked on one instance will remain valid on others. This also creates a memory leak in a long-running process. Use a distributed cache like Redis for the revocation list to ensure state is shared across all instances.

cache/redis-client.js
```diff
@@ -0,0 +1,386 @@
...
+        });
+    }
+
+    async flush() {
+        if (!this.connected) {
+            await this.connect();
+        }
+
+        try {
+            await this.client.flushAll();
```
YashwanthQdex, line 329: **Issue in cache/redis-client.js at Line 329**
**Severity:** Critical

**Problematic Code:**
```diff
+            await this.client.flushAll();
```

**Comments:**
Dangerous Exposure of `FLUSHALL` Command

**Before Fix:**
```
async flush() {
    if (!this.connected) {
        await this.connect();
    }

    try {
        await this.client.flushAll();
        return true;
    } catch (error) {
        console.error('Redis FLUSH error:', error);
        return false;
    }
}
```

**After Fix:**
> // Method removed entirely.

**Recommendation:**
The `flush` method exposes the `FLUSHALL` command, which deletes all data from all databases in the Redis instance. This is an extremely destructive operation that should never be exposed through a general-purpose application client. A single accidental or malicious call could wipe out all cached data, sessions, and more. Remove this method entirely. Flushing should only be done via secure, administrative scripts with strict access controls.

api/http-client.js
```diff
@@ -0,0 +1,378 @@
...
+        const promises = requests.map(req =>
+            this.request(req.method, req.url, req.data, req.options)
+        );
+
+        return Promise.allSettled(promises);
+    }
+
+    async getWithCache(url, options = {}) {
+        // Simple in-memory cache
+        if (!this.cache) {
```
YashwanthQdex, line 215: **Issue in api/http-client.js at Line 215**
**Severity:** Medium

**Problematic Code:**
```diff
+        if (!this.cache) {
```

**Comments:**
Unbounded In-Memory Cache Causes Memory Leak

**Before Fix:**
```
if (!this.cache) {
    this.cache = new Map();
}
```

**After Fix:**
```
const LRUCache = require('lru-cache');
// In constructor:
this.cache = new LRUCache({ max: 500 }); // Set a reasonable limit
```

**Recommendation:**
The `getWithCache` method uses a standard `Map` for caching without any size limit or eviction policy (like LRU). In a long-running server process, this cache will grow indefinitely, consuming more and more memory, eventually leading to a crash. Replace the `Map` with a proper LRU cache implementation (e.g., using the `lru-cache` library) to enforce a memory limit.

cache/redis-client.js
```diff
@@ -0,0 +1,386 @@
...
+        }
+    }
+
+    async keys(pattern = '*') {
+        if (!this.connected) {
+            await this.connect();
+        }
+
+        try {
+            return await this.client.keys(pattern);
```
YashwanthQdex, line 152: **Issue in cache/redis-client.js at Line 152**
**Severity:** Medium

**Problematic Code:**
```diff
+            return await this.client.keys(pattern);
```

**Comments:**
Use of Blocking `KEYS` Command in Production

**Before Fix:**
```
return await this.client.keys(pattern);
```

**After Fix:**
```
const allKeys = [];
for await (const key of this.client.scanIterator({ MATCH: pattern, COUNT: 100 })) {
    allKeys.push(key);
}
return allKeys;
```

**Recommendation:**
The `keys` method uses the Redis `KEYS` command, which is a blocking operation that scans the entire keyspace. On a production database with many keys, this can block the single-threaded Redis server for seconds, causing performance degradation for all other clients. Replace `KEYS` with the non-blocking `SCAN` command, which iterates through keys using a cursor.

cache/redis-client.js
```diff
@@ -0,0 +1,386 @@
...
+    async get(key) {
+        if (!this.connected) {
+            await this.connect();
+        }
+
+        try {
+            const value = await this.client.get(key);
+            return value ? JSON.parse(value) : null;
+        } catch (error) {
+            console.error('Redis GET error:', error);
```
YashwanthQdex, line 86: **Issue in cache/redis-client.js at Line 86**
**Severity:** Medium

**Problematic Code:**
```diff
+            console.error('Redis GET error:', error);
```

**Comments:**
Error Masking in Cache Operations

**Before Fix:**
```
} catch (error) {
    console.error('Redis GET error:', error);
    return null;
}
```

**After Fix:**
```
} catch (error) {
    console.error('Redis GET error:', error);
    throw new Error('Failed to retrieve data from cache.');
}
```

**Recommendation:**
Methods like `get`, `set`, and `delete` catch errors, log them to the console, and return a default value (`null`, `false`, etc.). This masks the failure from the calling code. For example, `get` returning `null` is ambiguous—it could mean a cache miss or a Redis connection failure. This can lead to inconsistent state or stale data being used. Re-throw the error (or a custom application error) so the caller is forced to handle the failure explicitly.

auth/jwt-helper.js
```diff
@@ -0,0 +1,521 @@
...
+            iss: this.issuer,
+            aud: this.audience
+        };
+
+        const tokenOptions = {
+            expiresIn: options.expiresIn || this.expiresIn,
+            algorithm: this.algorithm
+        };
+
+        const token = jwt.sign(tokenPayload, this.secretKey, tokenOptions);
```
YashwanthQdex, line 30: **Issue in auth/jwt-helper.js at Line 30**
**Severity:** Medium

**Problematic Code:**
```diff
+        const token = jwt.sign(tokenPayload, this.secretKey, tokenOptions);
```

**Comments:**
Synchronous JWT Operations Can Block Event Loop

**Before Fix:**
```
const token = jwt.sign(tokenPayload, this.secretKey, tokenOptions);
```

**After Fix:**
```
const token = await new Promise((resolve, reject) => {
    jwt.sign(tokenPayload, this.secretKey, tokenOptions, (err, encoded) => {
        if (err) return reject(err);
        resolve(encoded);
    });
});
```

**Recommendation:**
The code uses synchronous methods like `jwt.sign` and `jwt.verify`. While convenient, these can block the Node.js event loop, especially under heavy load or with complex payloads, degrading server performance. Use the asynchronous, callback-based or promise-based versions of these methods provided by the `jsonwebtoken` library to avoid blocking.

api/http-client.js
```diff
@@ -0,0 +1,378 @@
...
+        }
+
+        return new Promise((resolve, reject) => {
+            const fileStream = fs.createWriteStream(destination);
+
+            fileStream.on('finish', () => resolve(destination));
+            fileStream.on('error', reject);
+
+            // Mock writing response data
+            fileStream.write(JSON.stringify(response.data));
```
YashwanthQdex, line 175: **Issue in api/http-client.js at Line 175**
**Severity:** Low

**Problematic Code:**
```diff
+            fileStream.write(JSON.stringify(response.data));
```

**Comments:**
Incorrect File Download Implementation

**Before Fix:**
```
// Mock writing response data
fileStream.write(JSON.stringify(response.data));
fileStream.end();
```

**After Fix:**
```
// This requires changing makeRequest to handle streams.
// Assuming `this.getStream(url, options)` returns the raw response stream:
const responseStream = await this.getStream(url, options);
responseStream.pipe(fileStream);
fileStream.on('finish', () => resolve(destination));
```

**Recommendation:**
The `download` method buffers the entire response in memory, stringifies it as JSON, and then writes it to a file. This is incorrect for two reasons: 1) It consumes excessive memory for large files. 2) It will corrupt any non-JSON or binary data. The correct approach is to stream the response body directly to the file. This requires modifying `makeRequest` to provide access to the response stream.

cache/redis-client.js
```diff
@@ -0,0 +1,386 @@
...
+        try {
+            await this.client.publish(channel, JSON.stringify(message));
+            return true;
+        } catch (error) {
+            console.error('Redis PUBLISH error:', error);
+            return false;
+        }
+    }
+
+    subscribe(channel, callback) {
```
YashwanthQdex, line 308: **Issue in cache/redis-client.js at Line 308**
**Severity:** Low

**Problematic Code:**
```diff
+    subscribe(channel, callback) {
```

**Comments:**
Subscribe Method Leaks Memory

**Before Fix:**
```
subscribe(channel, callback) {
    // ...
    this.client.subscribe(channel, (message) => {
        // ...
    });
}
```

**After Fix:**
```
subscribe(channel, callback) {
    // ...
    const subscriber = this.client.duplicate();
    subscriber.connect();
    subscriber.subscribe(channel, (message) => { ... });

    return {
        unsubscribe: () => subscriber.unsubscribe(channel).then(() => subscriber.quit())
    };
}
```

**Recommendation:**
The `subscribe` method sets up a listener but provides no way to `unsubscribe`. This means the callback and its closure will be held in memory indefinitely, creating a memory leak every time it's called. The method should return a function or an object that allows the caller to clean up the subscription.

cache/redis-client.js
```diff
@@ -0,0 +1,386 @@
...
+
+    async disconnect() {
+        if (this.client) {
+            await this.client.quit();
+            this.connected = false;
+        }
+    }
+
+    async set(key, value, ttl = null) {
+        if (!this.connected) {
```
YashwanthQdex, line 59: **Issue in cache/redis-client.js at Line 59**
**Severity:** Medium

**Problematic Code:**
```diff
added +        if (!this.connected) {
```

**Comments:**
Duplicated Connection Check in Every Method

**Before Fix:**
```
if (!this.connected) {
    await this.connect();
}
```

**After Fix:**
> // This block should be removed. Rely on the redis client's built-in command queueing and reconnection logic.

**Recommendation:**
The connection check `if (!this.connected) { await this.connect(); }` is repeated in almost every method. This violates the DRY principle and makes the code harder to maintain. A better approach is to rely on the `node-redis` library's built-in reconnection and command buffering capabilities, which handle this automatically. Remove these manual checks and configure the client's reconnection strategy properly in the `connect` method.

cache/redis-client.js
```diff
@@ -0,0 +1,386 @@
...
+        try {
+            const serialized = JSON.stringify(value);
+            if (ttl) {
+                await this.client.setEx(key, ttl, serialized);
+            } else {
+                await this.client.set(key, serialized);
+            }
+            return true;
+        } catch (error) {
+            console.error('Redis SET error:', error);
```
YashwanthQdex, line 72: **Issue in cache/redis-client.js at Line 72**
**Severity:** Medium

**Problematic Code:**
```diff
added +            console.error('Redis SET error:', error);
```

**Comments:**
Errors are Logged and Swallowed

**Before Fix:**
```
} catch (error) {
    console.error('Redis SET error:', error);
    return false;
}
```

**After Fix:**
```
} catch (error) {
    console.error(`Redis SET error for key '${key}':`, error);
    throw new Error(`Redis SET operation failed. Reason: ${error.message}`);
}
```

**Recommendation:**
Catching errors only to log them with `console.error` and returning a generic value (e.g., `false`, `null`) hides the root cause from the calling code. This makes debugging and proper error handling at the application level impossible. Instead of swallowing errors, re-throw them or throw a custom error with more context so the caller can decide how to handle the failure.

database/connection-pool.js
YashwanthQdex, line 473: **Issue in database/connection-pool.js at Line 473**
**Severity:** Critical

**Problematic Code:**
```diff
+        let sql = `CREATE USER '${username}'@'%' IDENTIFIED BY '${password}'`;
```

**Comments:**
Critical SQL Injection Vulnerability in User Creation

**Before Fix:**
```
let sql = `CREATE USER '${username}'@'%' IDENTIFIED BY '${password}'`;

await this.query(sql);

if (database) {
    await this.query(`GRANT ALL PRIVILEGES ON ${database}.* TO '${username}'@'%'`);
    await this.query('FLUSH PRIVILEGES');
}
```

**After Fix:**
```
// DDL statements like CREATE USER often don't support placeholders for names.
// Therefore, strict validation is essential.
const validUsernameRegex = /^[a-zA-Z0-9_]+$/;
if (!validUsernameRegex.test(username)) {
    throw new Error('Invalid username format.');
}

// Use placeholders for values like passwords.
let sql = `CREATE USER ?@'%' IDENTIFIED BY ?`;
await this.query(sql, [username, password]);

if (database) {
    // Also validate and escape database names if they can be user-influenced.
...
```

**Recommendation:**
The `username` and `password` variables are directly interpolated into the SQL string, which is a classic SQL injection vulnerability. An attacker could provide a malicious username or password to execute arbitrary SQL commands. Use parameterized queries for all values. For identifiers that cannot be parameterized, they must be strictly validated against a whitelist and/or escaped.

database/connection-pool.js
```diff
@@ -0,0 +1,488 @@
...
+        const params = [];
+
+        if (Object.keys(conditions).length > 0) {
+            const whereParts = Object.keys(conditions).map(key => `${key} = ?`);
+            sql += ` WHERE ${whereParts.join(' AND ')}`;
+            params.push(...Object.values(conditions));
+        }
+
+        if (options.orderBy) {
+            sql += ` ORDER BY ${options.orderBy}`;
```
YashwanthQdex, line 179: **Issue in database/connection-pool.js at Line 179**
**Severity:** Critical

**Problematic Code:**
```diff
+            sql += ` ORDER BY ${options.orderBy}`;
```

**Comments:**
SQL Injection Vulnerability in ORDER BY Clause

**Before Fix:**
```
if (options.orderBy) {
    sql += ` ORDER BY ${options.orderBy}`;
}
```

**After Fix:**
```
if (options.orderBy) {
    const allowedColumns = ['id', 'name', 'created_at']; // Example whitelist
    if (allowedColumns.includes(options.orderBy)) {
        sql += ` ORDER BY ${mysql.escapeId(options.orderBy)}`; // Also escape the identifier
    } else {
        throw new Error(`Invalid sort column: ${options.orderBy}`);
    }
}
```

**Recommendation:**
The `options.orderBy` value is directly concatenated into the SQL query. If this value comes from user input (e.g., a URL query parameter), it allows an attacker to inject arbitrary SQL. You must validate the `orderBy` value against a predefined whitelist of allowed column names for the specific table before using it in the query.

database/connection-pool.js
```diff
@@ -0,0 +1,488 @@
...
+    async execute(sql, params = []) {
+        return this.query(sql, params);
+    }
+
+    async insert(table, data) {
+        const columns = Object.keys(data);
+        const values = Object.values(data);
+        const placeholders = columns.map(() => '?').join(', ');
+
+        const sql = `INSERT INTO ${table} (${columns.join(', ')}) VALUES (${placeholders})`;
```
YashwanthQdex, line 139: **Issue in database/connection-pool.js at Line 139**
**Severity:** Critical

**Problematic Code:**
```diff
+        const sql = `INSERT INTO ${table} (${columns.join(', ')}) VALUES (${placeholders})`;
```

**Comments:**
SQL Injection Vulnerability via Unsanitized Table and Column Names

**Before Fix:**
```
const sql = `INSERT INTO ${table} (${columns.join(', ')}) VALUES (${placeholders})`;
```

**After Fix:**
```
const ALLOWED_TABLES = ['users', 'products']; // Example allow-list
if (!ALLOWED_TABLES.includes(table)) {
    throw new Error('Invalid table name');
}
// Also validate column names

const sql = `INSERT INTO ${table} (${columns.join(', ')}) VALUES (${placeholders})`;
```

**Recommendation:**
Do not construct queries with interpolated table or column names that originate from user input. If dynamic table/column names are required, validate them against a strict allow-list of known, safe names before incorporating them into a query.

database/connection-pool.js
```diff
@@ -0,0 +1,488 @@
+const mysql = require('mysql2/promise');
+
+class ConnectionPool {
+    constructor(options = {}) {
+        this.host = options.host || 'localhost';
+        this.port = options.port || 3306;
+        this.user = options.user || 'root';
+        this.password = options.password || '';
```
YashwanthQdex, line 8: **Issue in database/connection-pool.js at Line 8**
**Severity:** Critical

**Problematic Code:**
```diff
+        this.password = options.password || '';
```

**Comments:**
Hardcoded Default Database Credentials

**Before Fix:**
```
this.user = options.user || 'root';
this.password = options.password || '';
```

**After Fix:**
```
if (!options.user || !options.password) {
    throw new Error('Database user and password are required.');
}
this.user = options.user;
this.password = options.password;
```

**Recommendation:**
Remove default values for sensitive options like `user` and `password`. Instead, throw an error if they are not provided to enforce explicit configuration. Credentials should be loaded securely from environment variables or a secrets manager.

files/file-manager.js
YashwanthQdex, line 32: **Issue in files/file-manager.js at Line 32**
**Severity:** Critical

**Problematic Code:**
```diff
+            fs.writeFileSync(fullPath, content, options.encoding || 'utf8');
```

**Comments:**
Blocking Event Loop with Synchronous File I/O

**Before Fix:**
```
fs.writeFileSync(fullPath, content, options.encoding || 'utf8');
```

**After Fix:**
```
const fs = require('fs').promises;
// ...
await fs.writeFile(fullPath, content, options.encoding || 'utf8');
```

**Recommendation:**
Replace all synchronous file system calls (e.g., `fs.writeFileSync`, `fs.readFileSync`) with their asynchronous, promise-based counterparts from `require('fs').promises`. This will prevent blocking the Node.js event loop, significantly improving application performance and scalability.

utils/logger.js
```diff
@@ -0,0 +1,222 @@
...
+        }
+    }
+
+    query(options = {}) {
+        try {
+            if (!fs.existsSync(this.logFile)) {
+                return [];
+            }
+
+            const content = fs.readFileSync(this.logFile, 'utf8');
```
YashwanthQdex, line 151: **Issue in utils/logger.js at Line 151**
**Severity:** Medium

**Problematic Code:**
```diff
+            const content = fs.readFileSync(this.logFile, 'utf8');
```

**Comments:**
Unscalable Log Querying Reads Entire File into Memory

**Before Fix:**
```
const content = fs.readFileSync(this.logFile, 'utf8');
const lines = content.trim().split('\n');
```

**After Fix:**
```
const readline = require('readline');
const stream = fs.createReadStream(this.logFile);
const rl = readline.createInterface({ input: stream, crlfDelay: Infinity });

for await (const line of rl) {
    // process line
}
```

**Recommendation:**
Refactor the `query` method to use streams to read the log file. This avoids loading the entire file into memory, which can cause high memory usage and application crashes with large log files. Use the `readline` module to process the file line-by-line.

files/file-manager.js
```diff
@@ -0,0 +1,494 @@
...
+
+    searchFiles(directory, searchTerm, options = {}) {
+        const files = this.listFiles(directory, options);
+        const results = [];
+
+        for (const file of files) {
+            if (file.isDirectory) continue;
+
+            try {
+                const content = this.readFile(file.path, { encoding: 'utf8' });
```
YashwanthQdex, line 345: **Issue in files/file-manager.js at Line 345**
**Severity:** Medium

**Problematic Code:**
```diff
+                const content = this.readFile(file.path, { encoding: 'utf8' });
```

**Comments:**
Potential Denial of Service by Reading Large Files into Memory

**Before Fix:**
```
const content = this.readFile(file.path, { encoding: 'utf8' });
if (content.includes(searchTerm)) {
```

**After Fix:**
```
// Refactor to use streams
const stream = fs.createReadStream(this.resolvePath(file.path), { encoding: 'utf8' });
for await (const chunk of stream) {
    if (chunk.includes(searchTerm)) {
        // Found a match, handle it and break if needed
    }
}
```

**Recommendation:**
Modify the `searchFiles` method to read files in chunks (streams) instead of loading the entire file content into memory. This mitigates the risk of a Denial of Service attack where an attacker could place a very large file in a searchable directory, causing the application to run out of memory.

config/manager.js
```diff
@@ -0,0 +1,277 @@
...
+        }
+    }
+
+    saveConfig() {
+        try {
+            const dir = path.dirname(this.configPath);
+            if (!fs.existsSync(dir)) {
+                fs.mkdirSync(dir, { recursive: true });
+            }
+            fs.writeFileSync(this.configPath, JSON.stringify(this.config, null, 2));
```
YashwanthQdex, line 50: **Issue in config/manager.js at Line 50**
**Severity:** Critical

**Problematic Code:**
```diff
+            fs.writeFileSync(this.configPath, JSON.stringify(this.config, null, 2));
```

**Comments:**
Insecurely Persisting Sensitive Configuration to Disk

**Before Fix:**
```
fs.writeFileSync(this.configPath, JSON.stringify(this.config, null, 2));
```

**After Fix:**
```
// Persisting configuration should be handled with care, ensuring no secrets are written.
const nonSensitiveConfig = { ...this.config };
delete nonSensitiveConfig.database.password; // Example of removing a secret
delete nonSensitiveConfig.redis.password;
fs.writeFileSync(this.configPath, JSON.stringify(nonSensitiveConfig, null, 2));
```

**Recommendation:**
Refactor the manager to never write secrets to disk. Configuration containing secrets (from environment variables or defaults) should exist only in memory. The JSON file should be for non-sensitive, shareable configuration only. Remove the `saveConfig()` calls from methods like `set()`, `update()`, and `applyEnvironmentOverrides()` or implement a filtering mechanism to exclude sensitive keys before writing.

config/manager.js
```diff
@@ -0,0 +1,277 @@
...
+            }
+        };
+        this.watchers = new Map();
+        this.loadConfig();
+    }
+
+    loadConfig() {
+        try {
+            if (fs.existsSync(this.configPath)) {
+                const data = fs.readFileSync(this.configPath, 'utf8');
```
YashwanthQdex, line 31: **Issue in config/manager.js at Line 31**
**Severity:** Critical

**Problematic Code:**
```diff
+                const data = fs.readFileSync(this.configPath, 'utf8');
```

**Comments:**
Use of Synchronous File I/O Blocks Event Loop

**Before Fix:**
```
loadConfig() {
    try {
        if (fs.existsSync(this.configPath)) {
            const data = fs.readFileSync(this.configPath, 'utf8');
            // ...
        }
    } catch (error) { /*...*/ }
}
```

**After Fix:**
```
async loadConfig() {
    try {
        // Note: existsSync is hard to replace async, better to handle error from readFile
        try {
            const data = await fs.promises.readFile(this.configPath, 'utf8');
            // ...
        } catch (e) { 
            if (e.code === 'ENOENT') { /* file does not exist */ } 
            else { throw e; }
        }
    } catch (error) { /*...*/ }
}
```

**Recommendation:**
Replace all synchronous file system calls (`readFileSync`, `writeFileSync`, `existsSync`, `mkdirSync`) with their asynchronous counterparts from the `fs.promises` API. This will prevent blocking the Node.js event loop, which is critical for the performance and responsiveness of a backend server.

cache/redis-client.js
YashwanthQdex, line 152: **Issue in cache/redis-client.js at Line 152**
**Severity:** Critical

**Problematic Code:**
```diff
+            return await this.client.keys(pattern);
```

**Comments:**
Use of Blocking Redis `KEYS` Command Risks Production Instability

**Before Fix:**
```
async keys(pattern = '*') {
    // ...
    try {
        return await this.client.keys(pattern);
    } catch (error) {
        // ...
    }
}
```

**After Fix:**
```
async keys(pattern = '*') {
    // ...
    try {
        const stream = this.client.scanIterator({ MATCH: pattern, COUNT: 100 });
        const keys = [];
        for await (const key of stream) {
            keys.push(key);
        }
        return keys;
    } catch (error) {
        // ...
    }
}
```

**Recommendation:**
The Redis `KEYS` command is a blocking operation that can stall the server while it scans the entire keyspace. It should not be used in production. Replace its usage in the `keys()` and `getKeyCount()` methods with the `SCAN` command, which provides a cursor-based iterator that does not block the server.

cache/redis-client.js
YashwanthQdex, line 59: **Issue in cache/redis-client.js at Line 59**
**Severity:** Medium

**Problematic Code:**
```diff
+        if (!this.connected) {
```

**Comments:**
Repetitive Connection Logic Violates DRY Principle

**Before Fix:**
```
async set(key, value, ttl = null) {
    if (!this.connected) {
        await this.connect();
    }
    // ...
}

async get(key) {
    if (!this.connected) {
        await this.connect();
    }
    // ...
}
```

**After Fix:**
```
async _ensureConnected() {
    if (!this.connected) {
        await this.connect();
    }
}

async set(key, value, ttl = null) {
    await this._ensureConnected();
    // ...
}

async get(key) {
    await this._ensureConnected();
    // ...
}
```

**Recommendation:**
Almost every method contains a duplicated block of code to check for and establish a connection (`if (!this.connected) { await this.connect(); }`). This should be extracted into a single private helper method, e.g., `_ensureConnected()`, which is then called at the start of each public method. This improves maintainability and reduces code duplication.

YashwanthQdex, line 59: **Issue in cache/redis-client.js at Line 59**
**Severity:** Medium

**Problematic Code:**
```diff
+        if (!this.connected) {
```

**Comments:**
Bug in `subscribe` Method Due to Missing `await`

**Before Fix:**
```
subscribe(channel, callback) {
    if (!this.connected) {
        this.connect();
    }

    this.client.subscribe(channel, (message) => {
        // ...
    });
}
```

**After Fix:**
```
async subscribe(channel, callback) {
    if (!this.connected) {
        await this.connect();
    }

    // Note: A client in subscriber mode can't issue other commands.
    // Consider using a dedicated client instance for subscriptions.
    this.client.subscribe(channel, (message) => {
        // ...
    });
}
```

**Recommendation:**
The `subscribe` method calls `this.connect()` without `await`. This creates a race condition where `this.client.subscribe` could be called before the connection is established, leading to an error. Convert the `subscribe` method to be `async` and `await` the connection call.

PR #44, opened 2025-12-02
Medium file pubsub test

YashwanthQdex:


logger/logger.go
```diff
@@ -0,0 +1,619 @@
...
+}
+
+func EnableColorOutput(enable bool) {
+	logMutex.Lock()
+	defer logMutex.Unlock()
+	enableColors = enable
+	defaultConfig.EnableColors = enable
+}
+
+func WithContext(key string, value interface{}) {
```
YashwanthQdex, line 182: **Issue in logger/logger.go at Line 182**
**Severity:** Critical

**Problematic Code:**
```diff
added +func WithContext(key string, value interface{}) {
```

**Comments:**
Global Context Leads to Race Conditions in Concurrent Environments

**Before Fix:**
```
	contextMutex.Lock()
	defer contextMutex.Unlock()
	contextFields[key] = value
}
```

**After Fix:**
```
// It is recommended to remove this global context pattern.
// Instead, pass context.Context to your logging functions.
// Example of a better function signature:
// func Info(ctx context.Context, msg string, args ...interface{})
```

**Recommendation:**
Remove the global context functionality (`contextFields`, `WithContext`, `WithTraceID`, etc.). Contextual information should be passed explicitly, ideally using Go's standard `context.Context`. Logging functions should accept a `context.Context` as their first argument, from which values like trace IDs can be extracted. This is the standard and safe way to handle request-scoped data in Go.

repository/repository.go
```diff
@@ -0,0 +1,698 @@
...
+)
+
+var (
+	stockHistory      []StockChange
+	reservations      []*models.Reservation
+	products          []models.Product
+	inventoryRecords  []models.InventoryRecord
+	stockMovements    []models.StockMovement
+	orders            []models.Order
+	mu                sync.Mutex
```
YashwanthQdex, line 21: **Issue in repository/repository.go at Line 21**
**Severity:** Critical

**Problematic Code:**
```diff
added +	mu                sync.Mutex
```

**Comments:**
Single Global Mutex Creates Severe Performance Bottleneck

**Before Fix:**
```
var (
    // ...
	mu                sync.Mutex
    // ...
)
```

**After Fix:**
```
var (
    // ...
	// Each data slice should have its own RWMutex for granular locking
	productsMu         sync.RWMutex
	products           map[string]models.Product // Use a map for efficient lookups

	inventoryRecordsMu sync.RWMutex
	inventoryRecords   map[string]models.InventoryRecord
    // ... and so on for other data types
)
```

**Recommendation:**
Replace the single global `sync.Mutex` with a more granular locking strategy. A `sync.RWMutex` should be used to allow concurrent reads. For better scalability, each data store (e.g., `products`, `inventoryRecords`) should have its own `RWMutex`. This will prevent unrelated operations (e.g., reading a product and writing an order) from blocking each other.

repository/repository.go
```diff
@@ -0,0 +1,698 @@
...
+	
+	reservations = append(reservations, res)
+	logger.Debug("Reservation saved", 
+		"id", res.ID,
+		"sku", res.SKU,
+		"quantity", res.Quantity,
+	)
+}
+
+func GetReservation(id string) (*models.Reservation, error) {
```
YashwanthQdex, line 113: **Issue in repository/repository.go at Line 113**
**Severity:** Critical

**Problematic Code:**
```diff
added +func GetReservation(id string) (*models.Reservation, error) {
```

**Comments:**
Data Race Vulnerability by Returning Pointers to Internal State

**Before Fix:**
```
func GetReservation(id string) (*models.Reservation, error) {
	mu.Lock()
	defer mu.Unlock()
	
		if res.ID == id {
			return res, nil
		}
	}
	return nil, errors.New("reservation not found")
}
```

**After Fix:**
```
func GetReservation(id string) (*models.Reservation, error) {
	mu.Lock()
	defer mu.Unlock()
	
		if res.ID == id {
            // Return a copy to prevent data races
			resCopy := *res 
			return &resCopy, nil
		}
	}
	return nil, errors.New("reservation not found")
}
```

**Recommendation:**
Functions that retrieve a single object must return a *copy* of the object, not a pointer to the internal version. This prevents the caller from causing data races by modifying the object after the repository's mutex has been released. Create a new instance of the struct, copy the data, and return a pointer to the new copy.

repository/repository.go
```diff
@@ -0,0 +1,698 @@
...
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	t.operations = append(t.operations, TransactionOperation{
+		Type:      opType,
+		Operation: operation,
+		Rollback:  rollback,
+	})
+}
+
+func (t *Transaction) Commit() error {
```
YashwanthQdex, line 470: **Issue in repository/repository.go at Line 470**
**Severity:** Medium

**Problematic Code:**
```diff
added +func (t *Transaction) Commit() error {
```

**Comments:**
Flawed Transaction Implementation Lacks Atomicity

**Before Fix:**
```
func (t *Transaction) Commit() error {
	t.mu.Lock()
	defer t.mu.Unlock()
	
	if t.committed || t.rolledBack {
		return errors.New("transaction already committed or rolled back")
	}
	
	for _, op := range t.operations {
		if err := op.Operation(); err != nil {
			// Rollback on error
			for i := len(t.operations) - 1; i >= 0; i-- {
				if t.operations[i].Rollback != nil {
					t.operations[i].Rollback()
				}
			}
			return fmt.Errorf("transaction failed at operation %s: %w", op.Type, err)
		}
	}
	...
```

**After Fix:**
```
// This transaction implementation is not atomic. It does not lock the shared repository data.
// A proper transaction would need to lock the repository for its entire duration.
// Example (conceptual):
// func (repo *Repository) ExecuteTransaction(fn func(tx *Transaction) error) error {
//     repo.mu.Lock() // Lock the entire repository
//     defer repo.mu.Unlock()
//
//     // Create a transactional copy of data
//     tx_data := repo.data.Copy()
//
//     err := fn(tx_data)
//     if err !=...
```

**Recommendation:**
The current transaction implementation is not atomic because it doesn't lock the repository for the duration of the transaction. Other goroutines can modify data mid-transaction. A correct implementation would require the transaction to hold a lock on the repository from `BeginTransaction` until `Commit` or `Rollback`. Given the complexity, consider if a simpler, explicit locking pattern in the service layer would suffice, or redesign the transaction to operate on a temporary state that is merged only on commit.

repository/repository.go
```diff
@@ -0,0 +1,698 @@
...
+			return nil
+		}
+	}
+	
+	products = append(products, product)
+	logger.Debug("Product saved", "sku", product.SKU)
+	return nil
+}
+
+func GetProduct(sku string) (*models.Product, error) {
```
YashwanthQdex, line 254: **Issue in repository/repository.go at Line 254**
**Severity:** Medium

**Problematic Code:**
```diff
added +func GetProduct(sku string) (*models.Product, error) {
```

**Comments:**
Inefficient Data Lookups with Linear Scans

**Before Fix:**
```
	mu.Lock()
	defer mu.Unlock()
	
	for _, p := range products {
		if p.SKU == sku {
			return &p, nil
		}
	}
}
```

**After Fix:**
```
// To improve performance, use a map for storage.
// var products map[string]models.Product

	mu.RLock() // Use RLock for reads
	defer mu.RUnlock()
	
	p, ok := products[sku]
	if !ok {
	}
	return &p, nil // Remember to return a copy here as well!
}
```

**Recommendation:**
All lookup operations (e.g., `GetProduct`, `GetInventoryRecord`) use a linear scan over a slice, which has O(n) time complexity and will be very slow with large datasets. Store the data in maps for efficient O(1) average-time lookups. For example, `products` should be a `map[string]models.Product` keyed by SKU.

models/models.go
```diff
@@ -0,0 +1,680 @@
...
+	"strings"
+	"time"
+)
+
+type Product struct {
+	SKU          string    `json:"sku"`
+	Name         string    `json:"name"`
+	Description  string    `json:"description,omitempty"`
+	Category     string    `json:"category,omitempty"`
+	Price        float64   `json:"price,omitempty"`
```
YashwanthQdex, line 16: **Issue in models/models.go at Line 16**
**Severity:** Medium

**Problematic Code:**
```diff
added +	Price        float64   `json:"price,omitempty"`
```

**Comments:**
Use of float64 for Monetary Values

**Before Fix:**
> Price        float64   `json:"price,omitempty"`
Cost         float64   `json:"cost,omitempty"`

**After Fix:**
```
// Using integers for cents is a common practice
PriceCents   int64   `json:"priceCents,omitempty"`
CostCents    int64   `json:"costCents,omitempty"`

// Or using a decimal library
// import "github.com/shopspring/decimal"
// Price        decimal.Decimal `json:"price,omitempty"`
// Cost         decimal.Decimal `json:"cost,omitempty"`
```

**Recommendation:**
Using `float64` for monetary values like `Price` and `Cost` is unsafe due to potential floating-point inaccuracies. Use an integer type to store the smallest currency unit (e.g., cents) or a dedicated decimal library (e.g., `github.com/shopspring/decimal`) to ensure precision in financial calculations.

models/models.go
```diff
@@ -0,0 +1,680 @@
+package models
```
YashwanthQdex, line 1: **Issue in models/models.go at Line 1**
**Severity:** Medium

**Problematic Code:**
```diff
added +package models
```

**Comments:**
Monolithic Models File Violates Single Responsibility Principle

**Before Fix:**
```
package models
```

**After Fix:**
```
// This file should be split into multiple files.
// For example, the Product struct and its methods should go into a new file `product.go`.

// file: product.go
package models

import (
    // ...
)

type Product struct {
    // ...
}

// ... Product methods
```

**Recommendation:**
The `models.go` file is over 600 lines and contains many distinct domain models (Product, Order, Supplier, etc.). This makes the code hard to navigate and maintain. Split the models into separate files based on their domain concept (e.g., `product.go`, `order.go`, `inventory.go`). This will significantly improve code organization and readability.

repository/repository.go
```diff
@@ -0,0 +1,698 @@
...
+	"inventory/models"
+	"strings"
+	"sync"
+	"time"
+)
+
+var (
+	stockHistory      []StockChange
+	reservations      []*models.Reservation
+	products          []models.Product
```
YashwanthQdex, line 17: **Issue in repository/repository.go at Line 17**
**Severity:** Medium

**Problematic Code:**
```diff
added +	products          []models.Product
```

**Comments:**
Use of Global Variables for Repository State

**Before Fix:**
```
var (
	stockHistory      []StockChange
	reservations      []*models.Reservation
	products          []models.Product
	inventoryRecords  []models.InventoryRecord
	stockMovements    []models.StockMovement
	orders            []models.Order
	mu                sync.Mutex
	repositoryConfig  RepositoryConfig
)
```

**After Fix:**
```
type Repository struct {
    mu sync.RWMutex
    products map[string]models.Product
    // ... other data stores
}

func NewRepository() *Repository {
    return &Repository{
        products: make(map[string]models.Product),
        // ...
    }
}
```

**Recommendation:**
The repository's state (products, orders, etc.) is stored in package-level global variables. This makes the code difficult to test in isolation and prevents creating multiple repository instances. Refactor the repository into a struct that holds the data slices and the mutex as fields. Provide a `NewRepository()` constructor function. This makes the component self-contained, easier to test, and more flexible.

models/models.go
```diff
@@ -0,0 +1,680 @@
...
+
+func (r *Reservation) CanFulfill() bool {
+	return r.IsValid()
+}
+
+func (r *Reservation) CanRelease() bool {
+	return r.Status == ReservationPending
+}
+
+func (r *Reservation) MarkFulfilled() {
```
YashwanthQdex, line 301: **Issue in models/models.go at Line 301**
**Severity:** Low

**Problematic Code:**
```diff
added +func (r *Reservation) MarkFulfilled() {
```

**Comments:**
Inconsistent Timestamps from Multiple `time.Now()` Calls

**Before Fix:**
```
func (r *Reservation) MarkFulfilled() {
	r.Status = ReservationFulfilled
	r.FulfilledAt = time.Now()
	r.UpdatedAt = time.Now()
	r.Version++
}
```

**After Fix:**
```
func (r *Reservation) MarkFulfilled() {
    now := time.Now()
	r.Status = ReservationFulfilled
	r.FulfilledAt = now
	r.UpdatedAt = now
	r.Version++
}
```

**Recommendation:**
Methods like `MarkFulfilled`, `Reserve`, and `Fulfill` call `time.Now()` multiple times to set different timestamp fields (e.g., `UpdatedAt`, `FulfilledAt`). This can result in slightly different values. It's better practice to call `time.Now()` once at the beginning of the method and reuse the resulting `time.Time` value for all timestamps set within that operation.

models/models.go
```diff
@@ -0,0 +1,680 @@
...
+
+func (o *Order) GetItemCount() int {
+	count := 0
+	for _, item := range o.Items {
+		count += item.Quantity
+	}
+	return count
+}
+
+func (o *Order) IsShippable() bool {
```
YashwanthQdex, line 505: **Issue in models/models.go at Line 505**
**Severity:** Low

**Problematic Code:**
```diff
added +func (o *Order) IsShippable() bool {
```

**Comments:**
Hardcoded String Literals for Order Status

**Before Fix:**
```
func (o *Order) IsShippable() bool {
	return o.Status == "CONFIRMED" || o.Status == "PROCESSING"
}
```

**After Fix:**
```
const (
    OrderStatusConfirmed = "CONFIRMED"
    OrderStatusProcessing = "PROCESSING"
    // ... other statuses
)

func (o *Order) IsShippable() bool {
	return o.Status == OrderStatusConfirmed || o.Status == OrderStatusProcessing
}
```

**Recommendation:**
The method `IsShippable` uses hardcoded strings `"CONFIRMED"` and `"PROCESSING"`. This is error-prone and makes the code harder to maintain. Define these statuses as constants, similar to how `ReservationStatus` is handled.

logger/logger.go
```diff
@@ -0,0 +1,619 @@
...
+	Level:        LevelInfo,
+	LogDir:       "./logs",
+	MaxFileSize:  10 * 1024 * 1024,
+	MaxBackups:   5,
+	OutputFormat: "text",
+	EnableColors: false,
+	EnableFileLog: false,
+}
+
+func init() {
```
YashwanthQdex, line 71: **Issue in logger/logger.go at Line 71**
**Severity:** Low

**Problematic Code:**
```diff
added +func init() {
```

**Comments:**
init() Function with I/O Side Effects

**Before Fix:**
```
func init() {
	setupFileLogging()
}
```

**After Fix:**
```
// The init() function should be removed.
// Configuration and setup should be handled in an explicit function that can return an error.

func New(config LoggerConfig) (*Logger, error) {
    // ... setup logic here ...
    if err := setupFileLogging(config); err != nil {
        return nil, err
    }
    // ...
}
```

**Recommendation:**
The `init()` function calls `setupFileLogging()`, which performs file system operations (creating directories, opening files). `init()` functions should not have side effects like I/O because errors cannot be returned and handled gracefully. All initialization that can fail should be moved to an explicit function (e.g., `NewLogger` or `Configure`) that can return an error.

notifications/notifications.go
```diff
@@ -0,0 +1,600 @@
...
+
+func getEmailRecipient() string {
+	if config.EmailFrom != "" {
+		return config.EmailFrom
+	}
+	return "admin@inventory.com"
+}
+
+func getSMSRecipient() string {
+	return "+1234567890" // Default SMS recipient
```
YashwanthQdex, line 280: **Issue in notifications/notifications.go at Line 280**
**Severity:** Critical

**Problematic Code:**
```diff
added +	return "+1234567890" // Default SMS recipient
```

**Comments:**
Hardcoded Secret (Phone Number)

**Before Fix:**
```
}
```

**After Fix:**
```
	// This should be loaded from config, not hardcoded.
	return config.SMSRecipient
}
```

**Recommendation:**
The function `getSMSRecipient` returns a hardcoded phone number. This is a critical security and privacy risk. Sensitive data like phone numbers must not be hardcoded. Load this value from a secure configuration source (e.g., environment variables, secret manager).

notifications/notifications.go
```diff
@@ -0,0 +1,600 @@
...
+			ID:        generateNotificationID(),
+			Type:      NotificationTypeEmail,
+			Priority:  priority,
+			Status:    StatusPending,
+			Recipient: getEmailRecipient(),
+			Subject:   fmt.Sprintf("Low Stock Alert: %s", alert.SKU),
+			Message:   formatLowStockMessage(alert),
+			CreatedAt: time.Now(),
+		}
+		sendNotification(notif)
```
YashwanthQdex, line 136: **Issue in notifications/notifications.go at Line 136**
**Severity:** Critical

**Problematic Code:**
```diff
added +	sendNotification(notif)
```

**Comments:**
Incorrect Retry Logic Causes Data Corruption

**Before Fix:**
```
func retryNotification(notif Notification) {
	time.Sleep(config.RetryDelay)
	notif.RetryCount++
	sendNotification(notif)
}
```

**After Fix:**
```
func retryNotification(notificationID string) {
	mu.Lock()
	var notifToRetry *Notification
	for i := range notifications {
		if notifications[i].ID == notificationID {
			notifications[i].RetryCount++
			notifToRetry = &notifications[i]
			break
		}
	}
	mu.Unlock()

	if notifToRetry != nil {
		sendNotification(*notifToRetry) // Or a new function that doesn't re-append
	}
}
```

**Recommendation:**
The `retryNotification` function calls `sendNotification` with a copy of the failed notification. `sendNotification` then appends this copy as a *new* entry to the `notifications` slice. This is incorrect. It should update the state of the *original* notification. This bug leads to duplicated notification records and breaks the `MaxRetries` logic, potentially causing infinite retries. Refactor the retry mechanism to operate on the original notification, perhaps by passing its ID.

notifications/notifications.go
```diff
@@ -0,0 +1,600 @@
...
+
+func sendNotification(notif Notification) {
+	mu.Lock()
+	notifications = append(notifications, notif)
+	mu.Unlock()
+
+	var err error
+	switch notif.Type {
+	case NotificationTypeEmail:
+		err = SendEmailAlert(models.LowStockAlert{SKU: notif.Message})
```
**Severity:** Critical

**Problematic Code:**
```diff
added +		err = SendEmailAlert(models.LowStockAlert{SKU: notif.Message})
```

**Comments:**
Incorrect Data Mapping in Notification Dispatch

**Before Fix:**
```
```

**After Fix:**
```
// The original alert data should be available here.
// For example, by changing the signature of sendNotification to:
// sendNotification(notif Notification, alertData models.LowStockAlert)
err = SendEmailAlert(alertData)
```

**Recommendation:**
When dispatching email and SMS alerts, a new `models.LowStockAlert` is created by incorrectly assigning the entire formatted `notif.Message` string to the `SKU` field. This corrupts the data sent to the alert functions. You must pass the original `LowStockAlert` data or its relevant fields, not the formatted message.

notifications/notifications.go
```diff
@@ -0,0 +1,600 @@
...
+		err = SendWebhookAlert(notif)
+	case NotificationTypeSlack:
+		err = SendSlackAlert(notif)
+	default:
+		err = errors.New("unknown notification type")
+	}
+
+	mu.Lock()
+	defer mu.Unlock()
+	for i := range notifications {
```
YashwanthQdex, line 245: **Issue in notifications/notifications.go at Line 245**
**Severity:** Critical

**Problematic Code:**
```diff
added +	for i := range notifications {
```

**Comments:**
Inefficient Notification Status Update (O(N) Complexity)

**Before Fix:**
```
mu.Lock()
defer mu.Unlock()
for i := range notifications {
	if notifications[i].ID == notif.ID {
		// update status
		break
	}
}
```

**After Fix:**
```
mu.Lock()
// Get index when appending
notifications = append(notifications, notif)
index := len(notifications) - 1
mu.Unlock()

// ... sending logic ...

mu.Lock()
defer mu.Unlock()
// Update directly by index
if index < len(notifications) && notifications[index].ID == notif.ID {
    // update status of notifications[index]
}
```

**Recommendation:**
The `sendNotification` function iterates through the entire `notifications` slice (which grows indefinitely) to find and update the status of the notification it just sent. This O(N) operation will become a severe performance bottleneck. Instead of searching, use the index of the newly added notification or a map for O(1) lookups.

notifications/notifications.go
```diff
@@ -0,0 +1,600 @@
...
+	cutoff := time.Now().Add(-olderThan)
+	var filtered []Notification
+	for _, n := range notifications {
+		if n.CreatedAt.After(cutoff) {
+			filtered = append(filtered, n)
+		}
+	}
+	notifications = filtered
+	logger.Info("Cleaned up old notifications", 
+		"removed", len(notifications)-len(filtered),
```
YashwanthQdex, line 460: **Issue in notifications/notifications.go at Line 460**
**Severity:** Critical

**Problematic Code:**
```diff
added +		"removed", len(notifications)-len(filtered),
```

**Comments:**
Bug in Cleanup Logging Logic

**Before Fix:**
```
notifications = filtered
logger.Info("Cleaned up old notifications", 
	"removed", len(notifications)-len(filtered),
	"remaining", len(filtered),
)
```

**After Fix:**
```
removedCount := len(notifications) - len(filtered)
notifications = filtered
logger.Info("Cleaned up old notifications", 
	"removed", removedCount,
	"remaining", len(notifications),
)
```

**Recommendation:**
The log message in `CleanupOldNotifications` calculates the number of removed items *after* the `notifications` slice has been reassigned. As a result, `len(notifications)-len(filtered)` is always zero, making the log misleading. Calculate the number of removed items before reassigning the slice.

notifications/notifications.go
```diff
@@ -0,0 +1,600 @@
...
+func GetNotificationMetrics() map[string]interface{} {
+	stats := GetNotificationStats()
+	
+	mu.Lock()
+	defer mu.Unlock()
+	
+	metrics := map[string]interface{}{
+		"total_processed": len(notifications),
+		"queue_size":      len(notificationQueue),
+		"subscribers":     len(subscribers),
```
YashwanthQdex, line 592: **Issue in notifications/notifications.go at Line 592**
**Severity:** Critical

**Problematic Code:**
```diff
added +		"subscribers":     len(subscribers),
```

**Comments:**
Race Condition on Subscribers Count

**Before Fix:**
```
metrics := map[string]interface{}{
    "subscribers":     len(subscribers),
}
```

**After Fix:**
```
subscriberMutex.RLock()
metrics := map[string]interface{}{
    "subscribers":     len(subscribers),
}
subscriberMutex.RUnlock()
```

**Recommendation:**
The function `GetNotificationMetrics` reads `len(subscribers)` without holding the `subscriberMutex`. Other goroutines can modify the `subscribers` slice concurrently (via `RegisterSubscriber` or `UnregisterSubscriber`), leading to a race condition. You must acquire a read lock on `subscriberMutex` before accessing the `subscribers` slice.

notifications/notifications.go
```diff
@@ -0,0 +1,600 @@
...
+	WebhookURL      string
+	SlackEnabled    bool
+	SlackWebhookURL string
+	MaxRetries      int
+	RetryDelay      time.Duration
+}
+
+var (
+	notificationQueue []models.LowStockAlert
+	notifications     []Notification
```
YashwanthQdex, line 77: **Issue in notifications/notifications.go at Line 77**
**Severity:** Medium

**Problematic Code:**
```diff
added +	notifications     []Notification
```

**Comments:**
Potential Memory Leak from Unbounded Slice

**Before Fix:**
```
var (
	notifications     []Notification
)
```

**After Fix:**
```
var (
	notifications     []Notification
)

// In main or an init function:
go func() {
    ticker := time.NewTicker(1 * time.Hour)
    for range ticker.C {
        CleanupOldNotifications(24 * time.Hour)
    }
}()
```

**Recommendation:**
The global `notifications` slice stores a history of all notifications and grows indefinitely. While a `CleanupOldNotifications` function exists, there is no mechanism shown that guarantees its periodic execution. This creates a high risk of a memory leak. Ensure a background job or cron is implemented to call this cleanup function regularly.

service.go
```diff
@@ -200,34 +173,31 @@ func (s *InventoryService) FulfillReservation(id string) (*Reservation, error) {
...
 	defer s.mu.RUnlock()
-	alerts := make([]LowStockAlert, 0)
+	alerts := make([]models.LowStockAlert, 0)
 	for sku, rec := range s.inventory {
 		available := rec.OnHand - rec.Reserved
 		if available <= s.config.LowStockThreshold {
-			alerts = append(alerts, LowStockAlert{SKU: sku, Available: available, Threshold: s.config.LowStockThreshold})
+			alert := models.LowStockAlert{SKU: sku, Available: available, Threshold: s.config.LowStockThreshold}
+			alerts = append(alerts, alert)
+			notifications.SendLowStockAlert(alert)
```
**Severity:** Medium

**Problematic Code:**
```diff
added +			notifications.SendLowStockAlert(alert)
```

**Comments:**
Risk of Notification Spam

**Before Fix:**
```
alerts = append(alerts, alert)
```

**After Fix:**
```
// Before sending, check if an alert for this SKU was sent recently.
if !wasAlertSentRecently(alert.SKU) {
    markAlertAsSent(alert.SKU)
}
```

**Recommendation:**
The `LowStockAlerts` function triggers a notification every time it is called for every low-stock item. If this function is run on a schedule, it will repeatedly send the same alerts, causing notification spam. Implement a deduplication or throttling mechanism to prevent sending the same alert for a given SKU within a defined time window.

notifications/notifications.go
```diff
@@ -0,0 +1,600 @@
...
+	}
+}
+
+func retryNotification(notif Notification) {
+	time.Sleep(config.RetryDelay)
+	notif.RetryCount++
+	sendNotification(notif)
+}
+
+func generateNotificationID() string {
```
YashwanthQdex, line 268: **Issue in notifications/notifications.go at Line 268**
**Severity:** Medium

**Problematic Code:**
```diff
added +func generateNotificationID() string {
```

**Comments:**
Non-Unique ID Generation

**Before Fix:**
```
func generateNotificationID() string {
	return fmt.Sprintf("notif-%d", time.Now().UnixNano())
}
```

**After Fix:**
```
import "github.com/google/uuid"

func generateNotificationID() string {
	return uuid.New().String()
}
```

**Recommendation:**
Notification IDs are generated using `time.Now().UnixNano()`. This is not guaranteed to be unique and can cause collisions in a high-throughput system. Use a more robust method for generating unique IDs, such as a UUID library (e.g., `github.com/google/uuid`).

notifications/notifications.go
```diff
@@ -0,0 +1,600 @@
...
+	SlackWebhookURL string
+	MaxRetries      int
+	RetryDelay      time.Duration
+}
+
+var (
+	notificationQueue []models.LowStockAlert
+	notifications     []Notification
+	config            NotificationConfig
+	mu                sync.Mutex
```
YashwanthQdex, line 79: **Issue in notifications/notifications.go at Line 79**
**Severity:** Medium

**Problematic Code:**
```diff
added +	mu                sync.Mutex
```

**Comments:**
Global State and Coarse-Grained Locking

**Before Fix:**
```
var (
    ...
    mu sync.Mutex
)
```

**After Fix:**
```
type NotificationService struct {
    mu sync.RWMutex
    config NotificationConfig
    notifications []Notification
    // ... other fields
}

func NewNotificationService(cfg NotificationConfig) *NotificationService { ... }
```

**Recommendation:**
The package relies on global variables (`notifications`, `config`, etc.) protected by a single `sync.Mutex`. This singleton pattern makes the package difficult to test and can cause performance bottlenecks. Refactor to use a struct (e.g., `NotificationService`) to encapsulate state and methods. This allows for better testability and dependency injection. Also, consider using `sync.RWMutex` to allow concurrent reads.

service.go
```diff
@@ -24,118 +21,92 @@ var (
...
 	return *rec, nil
 }
 
-<<<<<<< HEAD
-func (s *InventoryService) ReserveStock(sku string, qty int) (*Reservation, error) {
+func (s *InventoryService) ReserveStock(sku string, qty int) (*models.Reservation, error) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
+	if !validator.ValidateSKU(sku) || !validator.ValidateQuantity(qty) {
+		return nil, ErrProductNotFound
```
YashwanthQdex, line 108: **Issue in service.go at Line 108**
**Severity:** Medium

**Problematic Code:**
```diff
added +		return nil, ErrProductNotFound
```

**Comments:**
Misleading Error on Validation Failure

**Before Fix:**
```
if !validator.ValidateSKU(sku) || !validator.ValidateQuantity(qty) {
    return nil, ErrProductNotFound
}
```

**After Fix:**
```
var ErrInvalidQuantity = errors.New("invalid quantity")

// ...

if !validator.ValidateSKU(sku) {
    return nil, ErrProductNotFound
}
if !validator.ValidateQuantity(qty) {
    return nil, ErrInvalidQuantity
}
```

**Recommendation:**
In `ReserveStock`, if the quantity validation `!validator.ValidateQuantity(qty)` fails, the function returns `ErrProductNotFound`. This is misleading. The product exists, but the input quantity is invalid. Return a more appropriate error, such as a new `ErrInvalidQuantity`.

notifications/notifications.go
```diff
@@ -0,0 +1,600 @@
...
+	)
+}
+
+func formatLowStockSMS(alert models.LowStockAlert) string {
+	return fmt.Sprintf("LOW STOCK: %s - %d available (threshold: %d)", 
+		alert.SKU, alert.Available, alert.Threshold)
+}
+
+func formatLowStockJSON(alert models.LowStockAlert) string {
+	data, _ := json.Marshal(alert)
```
YashwanthQdex, line 210: **Issue in notifications/notifications.go at Line 210**
**Severity:** Low

**Problematic Code:**
```diff
added +	data, _ := json.Marshal(alert)
```

**Comments:**
Ignored Error

**Before Fix:**
```
data, _ := json.Marshal(alert)
```

**After Fix:**
```
data, err := json.Marshal(alert)
if err != nil {
    logger.Error("Failed to marshal low stock alert to JSON", err, "sku", alert.SKU)
    return "{}" // or some other fallback
}
```

**Recommendation:**
The error returned by `json.Marshal` in `formatLowStockJSON` is ignored. While marshalling this struct is unlikely to fail, it is a best practice to always handle or log errors to prevent silent failures.

notifications/notifications.go
```diff
@@ -0,0 +1,600 @@
...
+	// Process alert immediately
+	go processLowStockAlert(alert)
+}
+
+func processLowStockAlert(alert models.LowStockAlert) {
+	priority := determinePriority(alert)
+	
+	// Create notification for each enabled channel
+	if config.EmailEnabled {
+		notif := Notification{
```
YashwanthQdex, line 126: **Issue in notifications/notifications.go at Line 126**
**Severity:** Low

**Problematic Code:**
```diff
added +		notif := Notification{
```

**Comments:**
Code Duplication

**Before Fix:**
```
if config.EmailEnabled {
    notif := Notification{...}
    sendNotification(notif)
}
if config.SMSEnabled ... {
    notif := Notification{...}
    sendNotification(notif)
}
```

**After Fix:**
```
func newLowStockNotification(typ NotificationType, recipient, message, subject string, priority NotificationPriority) Notification { ... }

// in processLowStockAlert:
if config.EmailEnabled {
    sendNotification(newLowStockNotification(...))
}
```

**Recommendation:**
The `processLowStockAlert` function contains several blocks of nearly identical code for creating `Notification` structs for different channels (Email, SMS, Webhook, Slack). This can be refactored into a helper function to reduce duplication and improve maintainability.

middleware/middleware.go
```diff
@@ -0,0 +1,635 @@
...
+			return
+		}
+
+		apiKey := r.Header.Get("X-API-Key")
+		if apiKey == "" {
+			apiKey = r.URL.Query().Get("api_key")
+		}
+
+		expectedKey := defaultMiddlewareConfig.APIKey
+		if expectedKey != "" && subtle.ConstantTimeCompare([]byte(apiKey), []byte(expectedKey)) != 1 {
```
YashwanthQdex, line 159: **Issue in middleware/middleware.go at Line 159**
**Severity:** Critical

**Problematic Code:**
```diff
added +		if expectedKey != "" && subtle.ConstantTimeCompare([]byte(apiKey), []byte(expectedKey)) != 1 {
```

**Comments:**
Authentication Bypass Vulnerability

**Before Fix:**
```
if expectedKey != "" && subtle.ConstantTimeCompare([]byte(apiKey), []byte(expectedKey)) != 1 {
```

**After Fix:**
```
expectedKey := defaultMiddlewareConfig.APIKey
		if expectedKey == "" {
			logger.Error("API key is required but not configured", "path", r.URL.Path)
			w.WriteHeader(http.StatusUnauthorized)
			// ... write error response
			return
		}

		if subtle.ConstantTimeCompare([]byte(apiKey), []byte(expectedKey)) != 1 {
```

**Recommendation:**
If `APIKeyRequired` is true, the middleware must fail-closed. Modify the logic to ensure that if the configured `APIKey` is empty, an 'Unauthorized' error is returned. This prevents requests from being processed without authentication when the key is not configured.

middleware/middleware.go
```diff
@@ -0,0 +1,635 @@
...
+}
+
+func (rw *responseWriter) Write(b []byte) (int, error) {
+	if !rw.wroteHeader {
+		rw.WriteHeader(http.StatusOK)
+	}
+	if rw.body == nil {
+		rw.body = &bytes.Buffer{}
+	}
+	rw.body.Write(b)
```
YashwanthQdex, line 40: **Issue in middleware/middleware.go at Line 40**
**Severity:** Critical

**Problematic Code:**
```diff
added +	rw.body.Write(b)
```

**Comments:**
Denial of Service (DoS) via Memory Exhaustion

**Before Fix:**
```
rw.body.Write(b)
```

**After Fix:**
```
// In a new response writer for metrics:

func (rw *metricsResponseWriter) Write(b []byte) (int, error) {
    n, err := rw.ResponseWriter.Write(b)
    rw.bytesWritten += n
    return n, err
}
```

**Recommendation:**
Avoid buffering the entire response body in memory. To measure response size for metrics, wrap the `ResponseWriter` and increment a counter within the `Write` method without storing the bytes. For logging, only buffer and log a small, truncated portion of the response body to prevent OOM errors with large responses.

middleware/middleware.go
```diff
@@ -0,0 +1,635 @@
...
+func (gz *gzipWriter) Write(b []byte) (int, error) {
+	if gz.writer == nil {
+		gz.Header().Set("Content-Encoding", "gzip")
+		gz.Header().Set("Vary", "Accept-Encoding")
+		gz.writer = io.MultiWriter(gz.ResponseWriter, nil) // Would use gzip.NewWriter in real impl
+	}
+	return gz.writer.Write(b)
+}
+
+func CompressionMiddleware(next http.Handler) http.Handler {
```
YashwanthQdex, line 593: **Issue in middleware/middleware.go at Line 593**
**Severity:** Critical

**Problematic Code:**
```diff
added +func CompressionMiddleware(next http.Handler) http.Handler {
```

**Comments:**
Broken Gzip Compression Middleware

**Before Fix:**
```
func CompressionMiddleware(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if !defaultMiddlewareConfig.EnableCompression {
			next.ServeHTTP(w, r)
			return
		}

		if !strings.Contains(r.Header.Get("Accept-Encoding"), "gzip") {
			next.ServeHTTP(w, r)
			return
		}

		next.ServeHTTP(w, r)
	})
}
```

**After Fix:**
```
import "compress/gzip"

// ... (inside middleware)
	gz, _ := gzip.NewWriter(w)
	defer gz.Close()
	next.ServeHTTP(gzipResponseWriter{ResponseWriter: w, Writer: gz}, r)
```

**Recommendation:**
The `CompressionMiddleware` is non-functional and will break clients expecting gzip. It sets the `Content-Encoding` header but does not compress the response. Replace this with a standard, working implementation like `github.com/gorilla/handlers.CompressHandler` or correctly implement it by wrapping the `http.ResponseWriter` with a `gzip.Writer`.

middleware/middleware.go
```diff
@@ -0,0 +1,635 @@
...
+				})
+				return
+			}
+			r.Body = http.MaxBytesReader(w, r.Body, maxSize)
+			next.ServeHTTP(w, r)
+		})
+	}
+}
+
+func TimeoutMiddleware(timeout time.Duration) func(http.Handler) http.Handler {
```
YashwanthQdex, line 352: **Issue in middleware/middleware.go at Line 352**
**Severity:** Medium

**Problematic Code:**
```diff
added +func TimeoutMiddleware(timeout time.Duration) func(http.Handler) http.Handler {
```

**Comments:**
Resource Leak in Timeout Middleware

**Before Fix:**
```
go func() {
	next.ServeHTTP(w, r)
	done <- true
}()

select {
case <-done:
	return
case <-ctx.Done():
    // ... write timeout error
}
```

**After Fix:**
```
func TimeoutMiddleware(timeout time.Duration) func(http.Handler) http.Handler {
    return func(next http.Handler) http.Handler {
        return http.TimeoutHandler(next, timeout, "{\"error\":\"Request timeout\"}")
    }
}
```

**Recommendation:**
The `TimeoutMiddleware` does not stop the handler's execution on timeout, leading to resource leaks. Replace the custom implementation with the standard library's `http.TimeoutHandler`, which correctly cancels the request context and prevents background goroutines from continuing to run after a timeout.

validator/validator.go
```diff
@@ -0,0 +1,755 @@
...
+
+func ValidatePostalCode(postalCode, country string) error {
+	if strings.TrimSpace(postalCode) == "" {
+		return errors.New("postal code cannot be empty")
+	}
+	
+	// Basic validation - can be extended for specific countries
+	switch strings.ToUpper(country) {
+	case "US":
+		usPattern := regexp.MustCompile(`^\d{5}(-\d{4})?$`)
```
YashwanthQdex, line 297: **Issue in validator/validator.go at Line 297**
**Severity:** Medium

**Problematic Code:**
```diff
added +		usPattern := regexp.MustCompile(`^\d{5}(-\d{4})?$`)
```

**Comments:**
Inefficient On-the-fly Regex Compilation

**Before Fix:**
```
func ValidatePostalCode(postalCode, country string) error {
    // ...
    usPattern := regexp.MustCompile(`^\d{5}(-\d{4})?$`) // Compiles every time
    // ...
}
```

**After Fix:**
```

func ValidatePostalCode(postalCode, country string) error {
    // ...
    if !usPostalCodePattern.MatchString(postalCode) {
        // ...
    }
    // ...
}
```

**Recommendation:**
Regular expressions in `ValidatePostalCode` and other functions (`ValidateIPv6`, `ValidateUUID`, etc.) are compiled on every call, which is inefficient. Compile these regexes once at the package level and store them in global variables to improve performance.

middleware/middleware.go
```diff
@@ -0,0 +1,635 @@
...
+		for ip, v := range rl.visitors {
+			if now.Sub(v.lastSeen) > rl.window*2 {
+				delete(rl.visitors, ip)
+			}
+		}
+		rl.mu.Unlock()
+	}
+}
+
+func getClientIP(r *http.Request) string {
```
YashwanthQdex, line 277: **Issue in middleware/middleware.go at Line 277**
**Severity:** Medium

**Problematic Code:**
```diff
added +func getClientIP(r *http.Request) string {
```

**Comments:**
IP Spoofing Vulnerability in Rate Limiter

**Before Fix:**
```
func getClientIP(r *http.Request) string {
	ip := r.Header.Get("X-Forwarded-For")
	if ip != "" {
        // ... returns first IP
	}
	// ...
	return strings.Split(r.RemoteAddr, ":")[0]
}
```

**After Fix:**
```
// Add configuration for trusted proxies
var trustedProxies = []string{"127.0.0.1", "::1"}

func getClientIP(r *http.Request) string {
    remoteIP := net.ParseIP(strings.Split(r.RemoteAddr, ":")[0])
    isTrusted := false
    for _, trusted := range trustedProxies {
        if remoteIP.String() == trusted {
            isTrusted = true
            break
        }
    }

    if isTrusted {
        ip := r.Header.Get("X-Forwarded-For")
        if ip != "" {
            return strings.TrimSpace(st...
```

**Recommendation:**
The `getClientIP` function unconditionally trusts the `X-Forwarded-For` header, allowing attackers to bypass IP-based rate limiting by spoofing this header. Only trust this header if the request originates from a known, trusted proxy. Otherwise, fall back to using `r.RemoteAddr`.

middleware/middleware.go
```diff
@@ -0,0 +1,635 @@
...
+	rate     int
+	window   time.Duration
+}
+
+type visitor struct {
+	lastSeen time.Time
+	count    int
+}
+
+var globalRateLimiter = &rateLimiter{
```
YashwanthQdex, line 198: **Issue in middleware/middleware.go at Line 198**
**Severity:** Low

**Problematic Code:**
```diff
added +var globalRateLimiter = &rateLimiter{
```

**Comments:**
Unused Global Variable

**Before Fix:**
```
var globalRateLimiter = &rateLimiter{
	visitors: make(map[string]*visitor),
	rate:     100,
	window:   time.Minute,
}
```

**Recommendation:**
The global variable `globalRateLimiter` is declared but never used. This constitutes dead code and should be removed to improve code clarity and maintainability.

YashwanthQdex:
🤖 AI Code Analysis Summary
🔍 Classification: AI Authored (95% AI)

• The PR introduces multiple large, feature-complete, and highly generic packages (`logger`, `middleware`, `validator`, `notifications`, `repository`) in a single commit. This 'kitchen sink' approach is characteristic of AI generating an entire application skeleton from a high-level prompt.
• The `validator/validator.go` file is a massive library of generic validation functions for a wide array of types (email, IP, credit card, UUID, etc.). It's highly improbable a human would write this entire utility library from scratch for this specific PR; it strongly resembles AI-generated boilerplate.
• The code exhibits extreme mechanical consistency and repetitiveness across thousands of lines. For instance, the `models` and `repository` files contain dozens of perfectly structured, template-like methods and data models, lacking the typical variations and incremental design of human-authored code.

Code Review Results :

📊 Issues Found: 13 Critical, 13 Medium, 6 Low

---

PR #45, opened 2025-12-02
Medium file pubsub test

YashwanthQdex:


logger/logger.go
```diff
@@ -0,0 +1,619 @@
...
+}
+
+func EnableColorOutput(enable bool) {
+	logMutex.Lock()
+	defer logMutex.Unlock()
+	enableColors = enable
+	defaultConfig.EnableColors = enable
+}
+
+func WithContext(key string, value interface{}) {
```
YashwanthQdex, line 187: **Issue in logger/logger.go at Line 182**
**Severity:** Critical

**Problematic Code:**
```diff
added +func WithContext(key string, value interface{}) {
```

**Comments:**
Global Logger Context is Not Goroutine-Safe

**Before Fix:**
```
	contextMutex.Lock()
	defer contextMutex.Unlock()
	contextFields[key] = value
}
```

**After Fix:**
```
// Recommendation: Remove global context functions.
// Instead, use a logger instance with context.
// Example of a better approach:
// func (l *Logger) With(key string, value interface{}) *Logger {
// 	 newLogger := l.clone()
// 	 newLogger.fields[key] = value
// 	 return newLogger
// }
```

**Recommendation:**
The logger uses a global map (`contextFields`) for contextual data like `traceId`. In a concurrent web server, one request's context can overwrite another's, leading to incorrect logs. Instead of global context, pass a request-scoped logger instance (e.g., created in a middleware) or use `context.Context` to carry request-specific values.

service.go
```diff
@@ -200,34 +173,31 @@ func (s *InventoryService) FulfillReservation(id string) (*Reservation, error) {
...
 	defer s.mu.RUnlock()
-	alerts := make([]LowStockAlert, 0)
+	alerts := make([]models.LowStockAlert, 0)
 	for sku, rec := range s.inventory {
 		available := rec.OnHand - rec.Reserved
 		if available <= s.config.LowStockThreshold {
-			alerts = append(alerts, LowStockAlert{SKU: sku, Available: available, Threshold: s.config.LowStockThreshold})
+			alert := models.LowStockAlert{SKU: sku, Available: available, Threshold: s.config.LowStockThreshold}
+			alerts = append(alerts, alert)
+			notifications.SendLowStockAlert(alert)
```
**Severity:** Medium

**Problematic Code:**
```diff
added +			notifications.SendLowStockAlert(alert)
```

**Comments:**
Synchronous Blocking Call Inside a Loop

**Before Fix:**
```
	alert := models.LowStockAlert{SKU: sku, Available: available, Threshold: s.config.LowStockThreshold}
	alerts = append(alerts, alert)
}
```

**After Fix:**
```
alertsToSend := []models.LowStockAlert{}
        alert := models.LowStockAlert{SKU: sku, Available: available, Threshold: s.config.LowStockThreshold}
        alerts = append(alerts, alert)
        alertsToSend = append(alertsToSend, alert)
    }
}
s.mu.RUnlock() // Release lock before network calls

// Send notifications outside the lock and potentially in parallel
for ...
```

**Recommendation:**
The `notifications.SendLowStockAlert` function is called inside a loop while holding a read lock. If this is a synchronous network call, it can severely degrade performance and cause lock contention. Consider collecting all alerts and sending them in a batch after the loop, or dispatching notifications asynchronously in separate goroutines.

logger/logger.go
```diff
@@ -0,0 +1,619 @@
...
+
+const (
+	LevelDebug = "DEBUG"
+	LevelInfo  = "INFO"
+	LevelWarn  = "WARN"
+	LevelError = "ERROR"
+	LevelFatal = "FATAL"
+)
+
+var (
```
YashwanthQdex, line 24: **Issue in logger/logger.go at Line 24**
**Severity:** Medium

**Problematic Code:**
```diff
added +var (
```

**Comments:**
Extensive Use of Global State in Logger Package

**Before Fix:**
```
var (
	logLevel      = LevelInfo
	logFile       *os.File
	logFileWriter io.Writer
	logMutex      sync.Mutex
	contextFields = make(map[string]interface{})
	// ... and other global variables
)
```

**After Fix:**
```
type Logger struct {
	level      string
	output     io.Writer
	mu         sync.Mutex
	fields     map[string]interface{}
	// ... other config fields
}

func New(config LoggerConfig) *Logger {
	// ... initialization logic
}
```

**Recommendation:**
The logger is implemented as a singleton using global variables for configuration and state. This makes the code harder to test, prevents having different logger configurations for different modules, and creates hidden dependencies. Refactor the logger to be a struct that is instantiated in `main` and passed to components that need it via dependency injection.

logger/logger.go
```diff
@@ -0,0 +1,619 @@
...
+	if !defaultConfig.EnableFileLog {
+		return
+	}
+	if err := os.MkdirAll(logDir, 0755); err != nil {
+		log.Printf("Failed to create log directory: %v", err)
+		return
+	}
+
+	logPath := filepath.Join(logDir, fmt.Sprintf("inventory-%s.log", time.Now().Format("2006-01-02")))
+	file, err := os.OpenFile(logPath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
```
YashwanthQdex: **Issue in logger/logger.go at Line 85**
**Severity:** Medium

**Problematic Code:**
```diff
added +	file, err := os.OpenFile(logPath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
```

**Comments:**
Insecure Log File Permissions

**Before Fix:**
```
file, err := os.OpenFile(logPath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0666)
```

**After Fix:**
```
file, err := os.OpenFile(logPath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
```

**Recommendation:**
Log files are created with `0666` permissions, which are world-writable. This could allow any user on the system to tamper with the logs. Use more restrictive permissions, such as `0644` (owner read/write, others read-only), to enhance security.

logger/logger.go
```diff
@@ -0,0 +1,619 @@
...
+
+func init() {
+	setupFileLogging()
+}
+
+func setupFileLogging() {
+	if !defaultConfig.EnableFileLog {
+		return
+	}
+	if err := os.MkdirAll(logDir, 0755); err != nil {
```
YashwanthQdex, line 81: **Issue in logger/logger.go at Line 79**
**Severity:** Medium

**Problematic Code:**
```diff
added +	if err := os.MkdirAll(logDir, 0755); err != nil {
```

**Comments:**
Silent Failure on Log File Initialization

**Before Fix:**
```
func setupFileLogging() {
	if !defaultConfig.EnableFileLog {
		return
	}
	if err := os.MkdirAll(logDir, 0755); err != nil {
		log.Printf("Failed to create log directory: %v", err)
		return
	}
// ...
}
```

**After Fix:**
```
func setupFileLogging() error {
	if !defaultConfig.EnableFileLog {
		return nil
	}
	if err := os.MkdirAll(logDir, 0755); err != nil {
		return fmt.Errorf("failed to create log directory: %w", err)
	}
// ...
	return nil
}
```

**Recommendation:**
In `setupFileLogging`, errors during directory or file creation are logged to stdout, but the application continues without file logging. This can be misleading. The function should return an error, allowing the caller to decide whether to proceed or terminate if file logging is critical.

logger/logger.go
```diff
@@ -0,0 +1,619 @@
...
+
+func Warn(msg string, args ...interface{}) {
+	writeLog(LevelWarn, msg, args...)
+}
+
+func Debug(msg string, args ...interface{}) {
+	writeLog(LevelDebug, msg, args...)
+}
+
+func Error(msg string, err error) {
```
YashwanthQdex, line 425: **Issue in logger/logger.go at Line 418**
**Severity:** Medium

**Problematic Code:**
```diff
added +func Error(msg string, err error) {
```

**Comments:**
Error Logging Loses Valuable Context

**Before Fix:**
```
func Error(msg string, err error) {
	if err != nil {
		writeLog(LevelError, msg, "error", err.Error())
	} else {
		writeLog(LevelError, msg)
	}
}
```

**After Fix:**
```
func Error(msg string, err error) {
	if err != nil {
		// Using %+v preserves stack trace for wrapped errors
		writeLog(LevelError, msg, "error", fmt.Sprintf("%+v", err))
	} else {
		writeLog(LevelError, msg)
	}
}
```

**Recommendation:**
The `Error` function logs only the string representation of an error via `err.Error()`. This discards crucial debugging information like stack traces or wrapped error context. Use a logging method that preserves this information, for example by using `fmt.Sprintf("%+v", err)` or a structured logging library that can properly handle error objects.

repository/repository.go
```diff
@@ -0,0 +1,698 @@
...
+			return nil
+		}
+	}
+	
+	products = append(products, product)
+	logger.Debug("Product saved", "sku", product.SKU)
+	return nil
+}
+
+func GetProduct(sku string) (*models.Product, error) {
```
YashwanthQdex, line 254: **Issue in repository/repository.go at Line 254**
**Severity:** Critical

**Problematic Code:**
```diff
+func GetProduct(sku string) (*models.Product, error) {
```

**Comments:**
Inefficient Data Access using Linear Scan

**Before Fix:**
```
	mu.Lock()
	defer mu.Unlock()
	
	for _, p := range products {
		if p.SKU == sku {
			return &p, nil
		}
	}
}
```

**After Fix:**
```
// Assuming 'products' is now a map: var products map[string]models.Product
	mu.Lock()
	defer mu.Unlock()
	
	product, ok := products[sku]
	if !ok {
	}
	return &product, nil
}
```

**Recommendation:**
The current implementation uses slices for data storage and performs linear scans for lookups, updates, and deletions (O(n) complexity). This will perform poorly with a growing dataset. Use maps for direct, O(1) access by primary key (e.g., `map[string]models.Product` for products by SKU).

repository/repository.go
```diff
@@ -0,0 +1,698 @@
...
+	mu.Lock()
+	defer mu.Unlock()
+	
+	for _, p := range products {
+		if p.SKU == sku {
+			return &p, nil
```
YashwanthQdex, line 260: **Issue in repository/repository.go at Line 260**
**Severity:** Critical

**Problematic Code:**
```diff
+			return &p, nil
```

**Comments:**
Returning Pointer to Loop Variable Copy

**Before Fix:**
```
for _, p := range products {
    if p.SKU == sku {
        return &p, nil // BUG: returns pointer to a copy
    }
}
```

**After Fix:**
```
for i := range products {
    if products[i].SKU == sku {
        return &products[i], nil // CORRECT: returns pointer to element in slice
    }
}
```

**Recommendation:**
The code returns a pointer to `p`, which is a copy of the slice element created by the `for...range` loop. This is a common Go pitfall that leads to all returned pointers pointing to the same memory location containing the last element of the slice. To fix this, use an index-based loop and return a pointer to the actual slice element: `return &products[i], nil`.

repository/repository.go
```diff
@@ -0,0 +1,698 @@
...
+	
+	for _, res := range reservations {
+		if res.Status == models.ReservationPending && now.After(res.ExpiresAt) {
+			result = append(result, res)
+		}
+	}
+	return result
+}
+
+func CleanupExpiredReservations() int {
```
YashwanthQdex, line 642: **Issue in repository/repository.go at Line 642**
**Severity:** Critical

**Problematic Code:**
```diff
+func CleanupExpiredReservations() int {
```

**Comments:**
Deadlock due to Non-Reentrant Mutex Lock

**Before Fix:**
```
func CleanupExpiredReservations() int {
	expired := GetExpiredReservations() // First lock inside this function
	count := 0
	
	mu.Lock() // Second lock causes deadlock
	defer mu.Unlock()
	
	for _, res := range expired {
		res.MarkExpired()
		count++
	}
	
	logger.Info("Expired reservations cleaned up", "count", count)
	return count
}
```

**After Fix:**
```
func CleanupExpiredReservations() int {
	mu.Lock()
	defer mu.Unlock()

	var expired []*models.Reservation
	now := time.Now()
		if res.Status == models.ReservationPending && now.After(res.ExpiresAt) {
			expired = append(expired, res)
		}
	}

	count := 0
	for _, res := range expired {
		res.MarkExpired()
		count++
	}
	
	logger.Info("Expired reservations cleaned up", "count", count)
	return count
}
```

**Recommendation:**
The function `CleanupExpiredReservations` calls `GetExpiredReservations()`, which locks the global mutex `mu`. It then immediately tries to lock `mu` again. Since Go's `sync.Mutex` is not re-entrant, this will cause a deadlock. Refactor the logic to perform all operations within a single lock-unlock cycle.

validator/validator.go
```diff
@@ -0,0 +1,755 @@
...
+
+func ValidatePostalCode(postalCode, country string) error {
+	if strings.TrimSpace(postalCode) == "" {
+		return errors.New("postal code cannot be empty")
+	}
+	
+	// Basic validation - can be extended for specific countries
+	switch strings.ToUpper(country) {
+	case "US":
+		usPattern := regexp.MustCompile(`^\d{5}(-\d{4})?$`)
```
YashwanthQdex, line 297: **Issue in validator/validator.go at Line 297**
**Severity:** Medium

**Problematic Code:**
```diff
+		usPattern := regexp.MustCompile(`^\d{5}(-\d{4})?$`)
```

**Comments:**
Inefficient Regex Compilation in Validation Function

**Before Fix:**
```
func ValidatePostalCode(postalCode, country string) error {
    // ...
    usPattern := regexp.MustCompile(`^\d{5}(-\d{4})?$`)
    // ...
}
```

**After Fix:**
```

func ValidatePostalCode(postalCode, country string) error {
    // ...
    if !usPostalCodePattern.MatchString(postalCode) {
        // ...
    }
    // ...
}
```

**Recommendation:**
Regular expressions are compiled on every call to `ValidatePostalCode` and several other validation functions (`ValidateIPv6`, `ValidateUUID`, etc.). Regex compilation is a computationally expensive operation. These regexes should be compiled once at the package level using `regexp.MustCompile` and stored in global variables, just as `skuPattern` is.

validator/validator.go
```diff
@@ -0,0 +1,755 @@
...
+	for _, part := range parts {
+		num, err := strconv.Atoi(part)
+		if err != nil || num < 0 || num > 255 {
+			return false
+		}
+	}
+	return true
+}
+
+func ValidateIPv6(ip string) bool {
```
YashwanthQdex, line 362: **Issue in validator/validator.go at Line 362**
**Severity:** Medium

**Problematic Code:**
```diff
+func ValidateIPv6(ip string) bool {
```

**Comments:**
Incorrect IPv6 Validation

**Before Fix:**
```
func ValidateIPv6(ip string) bool {
	ipv6Pattern := regexp.MustCompile(`^([0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}$`)
	return ipv6Pattern.MatchString(ip)
}
```

**After Fix:**
```
import "net"

func ValidateIPv6(ip string) bool {
    parsedIP := net.ParseIP(ip)
    return parsedIP != nil && parsedIP.To4() == nil
}
```

**Recommendation:**
The regex for IPv6 validation is incomplete and does not handle common valid formats like compressed addresses (e.g., `::1`). The standard library's `net.ParseIP` function is much more robust and should be used to validate both IPv4 and IPv6 addresses.

models/models.go
```diff
@@ -0,0 +1,680 @@
...
+	return ir.CalculateAvailable() >= quantity
+}
+
+func (ir *InventoryRecord) Reserve(quantity int) error {
+	if !ir.CanReserve(quantity) {
+		return errors.New("insufficient available stock")
+	}
+	ir.Reserved += quantity
+	ir.Version++
+	ir.UpdatedAt = time.Now()
```
YashwanthQdex, line 199: **Issue in models/models.go at Line 199**
**Severity:** Medium

**Problematic Code:**
```diff
+	ir.UpdatedAt = time.Now()
```

**Comments:**
Hardcoded Time Dependency Hinders Testability

**Before Fix:**
```
func (ir *InventoryRecord) Reserve(quantity int) error {
    // ...
    ir.UpdatedAt = time.Now()
    return nil
}
```

**After Fix:**
```
// Option 1: Pass time as an argument
func (ir *InventoryRecord) Reserve(quantity int, now time.Time) error {
    // ...
    ir.UpdatedAt = now
    return nil
}

// Option 2: Use a Clock interface
type Clock interface {
    Now() time.Time
}

func (ir *InventoryRecord) Reserve(quantity int, clock Clock) error {
    // ...
    ir.UpdatedAt = clock.Now()
    return nil
}
```

**Recommendation:**
Methods like `Reserve`, `Release`, `MarkFulfilled`, etc., directly call `time.Now()`. This makes unit testing difficult because the current time cannot be controlled. To improve testability, inject a time source, either by passing `time.Time` as a parameter or by using a `Clock` interface that can be mocked.

repository/repository.go
```diff
@@ -0,0 +1,698 @@
...
+	"errors"
+	"fmt"
+	"inventory/logger"
+	"inventory/models"
+	"strings"
+	"sync"
+	"time"
+)
+
+var (
```
YashwanthQdex, line 14: **Issue in repository/repository.go at Line 14**
**Severity:** Medium

**Problematic Code:**
```diff
+var (
```

**Comments:**
Use of Global Singleton for Repository

**Before Fix:**
```
var (
	stockHistory      []StockChange
	// ... other global variables
	mu                sync.Mutex
)

func GetProduct(sku string) (*models.Product, error) { /* ... */ }
```

**After Fix:**
```
type ProductRepository interface {
    GetProduct(sku string) (*models.Product, error)
    // ... other methods
}

type InMemoryRepository struct {
    products map[string]models.Product
    mu       sync.RWMutex
}

func (r *InMemoryRepository) GetProduct(sku string) (*models.Product, error) { /* ... */ }
```

**Recommendation:**
The repository is implemented as a package-level singleton with global variables. This pattern makes the code difficult to test (due to shared global state) and inflexible. Refactor this into a struct (e.g., `InMemoryRepository`) that implements a repository interface. This allows for proper dependency injection and easier mocking in tests.

middleware/middleware.go
```diff
@@ -0,0 +1,635 @@
...
+
+				// Send error response
+				w.Header().Set("Content-Type", "application/json")
+				w.WriteHeader(http.StatusInternalServerError)
+				json.NewEncoder(w).Encode(map[string]string{
+					"error": "Internal server error",
+				})
+			}
+		}()
+		next.ServeHTTP(w, r)
```
YashwanthQdex, line 142: **Issue in middleware/middleware.go at Line 142**
**Severity:** Critical

**Problematic Code:**
```diff
added +		next.ServeHTTP(w, r)
```

**Comments:**
Broken Gzip Compression Middleware

**Before Fix:**
```
func CompressionMiddleware(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if !defaultMiddlewareConfig.EnableCompression {
			next.ServeHTTP(w, r)
			return
		}

		if !strings.Contains(r.Header.Get("Accept-Encoding"), "gzip") {
			next.ServeHTTP(w, r)
			return
		}

		next.ServeHTTP(w, r)
	})
}
```

**After Fix:**
```
import (
    // ...
    "compress/gzip"
)

func CompressionMiddleware(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if !defaultMiddlewareConfig.EnableCompression || !strings.Contains(r.Header.Get("Accept-Encoding"), "gzip") {
			next.ServeHTTP(w, r)
			return
		}

		gzw := gzip.NewWriter(w)
		defer gzw.Close()

		// Use a response writer that wra...
```

**Recommendation:**
Implement proper gzip compression using the `compress/gzip` package. The current implementation sets the `Content-Encoding: gzip` header but does not compress the response body, which will cause clients to fail when decoding the response.

notifications/notifications.go
```diff
@@ -0,0 +1,600 @@
...
+
+func ScheduleNotification(notif Notification, delay time.Duration) {
+	go func() {
+		time.Sleep(delay)
+		sendNotification(notif)
+	}()
+}
+
+func GetNotificationMetrics() map[string]interface{} {
+	stats := GetNotificationStats()
```
YashwanthQdex, line 584: **Issue in notifications/notifications.go at Line 584**
**Severity:** Critical

**Problematic Code:**
```diff
added +	stats := GetNotificationStats()
```

**Comments:**
Deadlock in GetNotificationMetrics

**Before Fix:**
```
func GetNotificationMetrics() map[string]interface{} {
	stats := GetNotificationStats()
	
	mu.Lock()
	defer mu.Unlock()
	// ...
}
```

**After Fix:**
```
func GetNotificationMetrics() map[string]interface{} {
	mu.Lock()
	defer mu.Unlock()

	stats := getNotificationStatsInternal() // Unlocked version
	
	metrics := map[string]interface{}{
		"total_processed": len(notifications),
		"queue_size":      len(notificationQueue),
		"subscribers":     len(subscribers),
	}
	
	for k, v := range stats {
		metrics[k] = v
	}
	
	return metrics
}

// getNotificationStatsInternal is an unlocked version for internal use
func getNotificationStatsInternal() map[strin...
```

**Recommendation:**
Refactor `GetNotificationMetrics` to avoid nested locking on the same mutex (`mu`). Create an internal, unlocked version of `GetNotificationStats` that can be called while the lock is already held to prevent the deadlock.

notifications/notifications.go
```diff
@@ -0,0 +1,600 @@
...
+
+func sendNotification(notif Notification) {
+	mu.Lock()
+	notifications = append(notifications, notif)
+	mu.Unlock()
+
+	var err error
+	switch notif.Type {
+	case NotificationTypeEmail:
+		err = SendEmailAlert(models.LowStockAlert{SKU: notif.Message})
```
**Severity:** Critical

**Problematic Code:**
```diff
added +		err = SendEmailAlert(models.LowStockAlert{SKU: notif.Message})
```

**Comments:**
Incorrect Business Logic in Notification Dispatch

**Before Fix:**
```
case NotificationTypeEmail:
```

**After Fix:**
```
// The Notification struct should carry the original alert data
// type Notification struct {
//     ...
//     AlertData models.LowStockAlert `json:"-"` // Example
// }

case NotificationTypeEmail:
	err = SendEmailAlert(notif.AlertData)
```

**Recommendation:**
The original structured `LowStockAlert` data should be passed to the sending functions (`SendEmailAlert`, `SendSMSAlert`). Reconstructing it by placing the entire formatted message into the `SKU` field is incorrect and will cause failures or corrupted alerts.

notifications/notifications.go
```diff
@@ -0,0 +1,600 @@
...
+
+	mu.Lock()
+	defer mu.Unlock()
+	for i := range notifications {
+		if notifications[i].ID == notif.ID {
+			if err != nil {
+				notifications[i].Status = StatusFailed
+				notifications[i].Error = err.Error()
+				if notifications[i].RetryCount < config.MaxRetries {
+					go retryNotification(notifications[i])
```
YashwanthQdex, line 251: **Issue in notifications/notifications.go at Line 251**
**Severity:** Medium

**Problematic Code:**
```diff
added +					go retryNotification(notifications[i])
```

**Comments:**
Flawed Notification Retry Logic

**Before Fix:**
```
if notifications[i].RetryCount < config.MaxRetries {
	go retryNotification(notifications[i])
}
```

**After Fix:**
```
if notifications[i].RetryCount < config.MaxRetries {
	// Pass a pointer to the existing notification to avoid creating duplicates
	go retryNotification(&notifications[i])
}
```

**Recommendation:**
The retry logic should update the original notification's state (e.g., `RetryCount`, `Status`) instead of creating a new notification entry for each retry attempt. Pass a pointer to the notification or its ID to the `retryNotification` function to modify it in place.

middleware/middleware.go
```diff
@@ -0,0 +1,635 @@
...
+		w.Header().Set("X-Request-ID", requestID)
+		logger.WithTraceID(requestID)
+		r = r.WithContext(context.WithValue(r.Context(), "request_id", requestID))
+
+		next.ServeHTTP(w, r)
+	})
+}
+
+func generateRequestID() string {
+	return fmt.Sprintf("%d-%d", time.Now().UnixNano(), len(fmt.Sprintf("%d", time.Now().UnixNano())))
```
YashwanthQdex, line 401: **Issue in middleware/middleware.go at Line 401**
**Severity:** Medium

**Problematic Code:**
```diff
added +	return fmt.Sprintf("%d-%d", time.Now().UnixNano(), len(fmt.Sprintf("%d", time.Now().UnixNano())))
```

**Comments:**
Weak Request ID Generation

**Before Fix:**
```
func generateRequestID() string {
	return fmt.Sprintf("%d-%d", time.Now().UnixNano(), len(fmt.Sprintf("%d", time.Now().UnixNano())))
}
```

**After Fix:**
```
import "github.com/google/uuid"

func generateRequestID() string {
	return uuid.New().String()
}
```

**Recommendation:**
Use a standard, robust method for generating unique IDs, such as a UUID library (e.g., `github.com/google/uuid`). The current timestamp-based approach is not guaranteed to be unique under high load and is not a standard format.

middleware/middleware.go
```diff
@@ -0,0 +1,635 @@
...
+}
+
+var globalRateLimiter = &rateLimiter{
+	visitors: make(map[string]*visitor),
+	rate:     100,
+	window:   time.Minute,
+}
+
+func RateLimitMiddleware(requestsPerMinute int) func(http.Handler) http.Handler {
+	rl := &rateLimiter{
```
YashwanthQdex, line 205: **Issue in middleware/middleware.go at Line 205**
**Severity:** Medium

**Problematic Code:**
```diff
added +	rl := &rateLimiter{
```

**Comments:**
Inefficient Per-Call Rate Limiter Instantiation

**Before Fix:**
```
func RateLimitMiddleware(requestsPerMinute int) func(http.Handler) http.Handler {
	rl := &rateLimiter{
		visitors: make(map[string]*visitor),
		rate:     requestsPerMinute,
		window:   time.Minute,
	}

	go rl.cleanupVisitors()

	return func(next http.Handler) http.Handler { // ...
```

**After Fix:**
```
// Use the already defined globalRateLimiter or a shared instance passed during setup.
var globalRateLimiter = NewRateLimiter(100, time.Minute)

func RateLimitMiddleware(rl *rateLimiter) func(http.Handler) http.Handler {
	return func(next http.Handler) http.Handler { // ...
```

**Recommendation:**
The `RateLimitMiddleware` should use a single, shared `rateLimiter` instance to enforce a consistent rate limit across the application. Creating a new instance on every call is inefficient and likely incorrect.

middleware/middleware.go
```diff
@@ -0,0 +1,635 @@
...
+		next.ServeHTTP(w, r)
+	})
+}
+
+func AuditMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
+		rw := &responseWriter{ResponseWriter: w, statusCode: http.StatusOK}
+
+		userID := r.Header.Get("X-User-ID")
```
YashwanthQdex, line 483: **Issue in middleware/middleware.go at Line 483**
**Severity:** Medium

**Problematic Code:**
```diff
added +		userID := r.Header.Get("X-User-ID")
```

**Comments:**
Insecure User Identification for Auditing

**Before Fix:**
```
userID := r.Header.Get("X-User-ID")
```

**After Fix:**
```
// Retrieve user ID from context after it has been set by auth middleware
var userID string
if id, ok := r.Context().Value("user_id").(string); ok {
    userID = id
} else {
    userID = "anonymous"
}
```

**Recommendation:**
Do not trust the `X-User-ID` header for audit logging as it can be spoofed. User identity should be verified by an authentication middleware and the verified ID should be passed securely via the request context.

notifications/notifications.go
```diff
@@ -0,0 +1,600 @@
...
+	WebhookEnabled  bool
+	WebhookURL      string
+	SlackEnabled    bool
+	SlackWebhookURL string
+	MaxRetries      int
+	RetryDelay      time.Duration
+}
+
+var (
+	notificationQueue []models.LowStockAlert
```
YashwanthQdex, line 76: **Issue in notifications/notifications.go at Line 76**
**Severity:** Medium

**Problematic Code:**
```diff
added +	notificationQueue []models.LowStockAlert
```

**Comments:**
Unreliable In-Memory Notification Storage

**Before Fix:**
```
var (
	notificationQueue []models.LowStockAlert
	notifications     []Notification
	// ...
)
```

**After Fix:**
```
// This package should be backed by a persistent store.
// For example, using an interface:
// type NotificationStore interface {
//     QueueAlert(alert models.LowStockAlert)
//     AddNotification(notif Notification)
//     // ... other methods
// }
```

**Recommendation:**
For a reliable system, use a persistent storage mechanism for the notification queue and history, such as a database or a dedicated message queue. In-memory storage will result in data loss on application restart.

notifications/notifications.go
```diff
@@ -0,0 +1,600 @@
...
+
+func getEmailRecipient() string {
+	if config.EmailFrom != "" {
+		return config.EmailFrom
+	}
+	return "admin@inventory.com"
+}
+
+func getSMSRecipient() string {
+	return "+1234567890" // Default SMS recipient
```
YashwanthQdex, line 280: **Issue in notifications/notifications.go at Line 280**
**Severity:** Medium

**Problematic Code:**
```diff
added +	return "+1234567890" // Default SMS recipient
```

**Comments:**
Hardcoded Configuration Value

**Before Fix:**
```
}
```

**After Fix:**
```
    if config.SMSRecipient != "" {
        return config.SMSRecipient
    }
	return "+1234567890" // Fallback default
}
```

**Recommendation:**
The SMS recipient phone number is hardcoded. This should be retrieved from the `NotificationConfig` struct to allow for proper configuration.

YashwanthQdex:
🤖 AI Code Analysis Summary
🔍 Classification: AI Authored (98% AI)

• The PR introduces multiple large, feature-complete packages (logger, middleware, models, notifications, repository, validator) in a single commit, which is characteristic of AI-driven scaffolding rather than incremental human development.
• The code exhibits extreme mechanical consistency in formatting, naming, and structure across all new files. For example, the repository functions all follow an identical pattern of locking, iterating, and unlocking.
• Many of the added modules, particularly `validator` and `middleware`, are overly comprehensive and generic, containing a vast library of functions (e.g., credit card validation, IPv6 validation, numerous security headers) that go far beyond the immediate needs of the project, suggesting generation from a generic template.

Code Review Results :

📊 Issues Found: 7 Critical, 15 Medium

---

logger/logger.go
YashwanthQdex: fixed - 7f8b78e2d4b6f24d5c055dc559619683b40f17ef

logger/logger.go
YashwanthQdex: fixed - 7f8b78e2d4b6f24d5c055dc559619683b40f17ef

logger/logger.go
YashwanthQdex: fixed - 7f8b78e2d4b6f24d5c055dc559619683b40f17ef

logger/logger.go
YashwanthQdex: fixed - 7f8b78e2d4b6f24d5c055dc559619683b40f17ef

logger/logger.go
YashwanthQdex: fixed - 7f8b78e2d4b6f24d5c055dc559619683b40f17ef

